PR #387, opened 2014-10-13
Clarify explanation of resources containment in classes.

mattpatey:
The documentation explaining the containment behaviour of a resource dependent on a class is unclear. It reads as though the dependent resource will be executed after each resource declared in the class has been fulfilled, which is incorrect.


jtappa:
We've been waiting for CLA for several months and contributor never commented back.

PR #428, opened 2015-01-15
(PDB-486) Added link for the new terminus failover docs

senior:
This is needed in connection with this pr: https://github.com/puppetlabs/puppetdb/pull/1212

### DIFF
--- a/source/_includes/puppetdb_master.html
+++ b/source/_includes/puppetdb_master.html
@@ -29,6 +29,7 @@
   <li><strong>Configuration</strong>
     <ul>
       <li><a href="/puppetdb/{{ puppetdb_version }}/configure.html">Configuring PuppetDB</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/puppetdb_connection.html">Configuring PuppetDB Connection</a></li>
       <li><a href="/puppetdb/{{ puppetdb_version }}/postgres_ssl.html">Setting Up SSL for PostgreSQL</a></li>
     </ul>
   <li><strong>Usage/Admin</strong>


senior:
@nfagerlund as I mentioned, we should probably also just point users from config_file_puppetdb.markdown to the new puppetdb_connection.markdown over on the PuppetDB side

source/_includes/puppetdb_master.html
```diff
@@ -29,6 +29,7 @@ <h2 id="puppetdb">PuppetDB NEXT</h2>
   <li><strong>Configuration</strong>
     <ul>
       <li><a href="/puppetdb/{{ puppetdb_version }}/configure.html">Configuring PuppetDB</a></li>
+      <li><a href="/puppetdb/{{ puppetdb_version }}/puppetdb_connection.html">Configuring PuppetDB Connection</a></li>
```
nfagerlund: What do you think about changing the title of the page? I suggest `puppetdb.conf: Configuring a Puppet/PuppetDB Connection`

Like, it's longer and uglier, but anyone can see exactly whether they need to care about that page immediately, and if someone is looking for puppetdb.conf specifically they won't have to even think about it.

senior: Fine with me, should I change it at the top of the PuppetDB PR as well?

nfagerlund: I think so, yeah. It's good for Google juice.

nfagerlund:
I'm going to put this change in a different PR along with edits to the puppetdb.conf page, so the puppetdb team can one-stop-shop when the main PR goes in.

PR #429, opened 2015-01-15
(PDB-486) puppetdb.conf page lives in PuppetDB docs now

nfagerlund:
This commit updates the Puppet reference manual page about that file, and adds
that page to the master version of the PuppetDB docs index.


nfagerlund:
PuppetDB team is free to merge this as soon as https://github.com/puppetlabs/puppetdb/pull/1212 goes in!

PR #431, opened 2015-01-22
command line option for puppet cert generate wrong

ApisMellow:
After hitting an issue with our puppet cert generate command, we figured out that the --dns_alt_names parameter uses underscores, not dashes. Switching from the given form to the underscore version solved our problem. I have a screenshot I can share if necessary that shows the incorrect version giving an error, switching the format and then having the same command succeed.

### DIFF
--- a/source/guides/install_puppet/post_install.markdown
+++ b/source/guides/install_puppet/post_install.markdown
@@ -60,7 +60,7 @@ This will create the CA certificate and the puppet master certificate, with the 
 
 You have two main options:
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
Dang, you're right. Thanks for catching that.

nfagerlund:
I acked the source to make sure, and this seems to be the only instance as applied to the `puppet cert` command. The `puppet certificate` and `puppet ca` faces are a different story, and they actually implement a separate `--dns-alt-names` option that they use instead of the auto-generated one from the Puppet setting. So we're good.

PR #432, opened 2015-01-27
(maint) Fix statement that says regular expressions cannot be assigned

hlindberg:
It is possible to assign a regular expression to a variable and use it
later. e.g.

$a = /ana/
notice "banana"=~ $a

Which will notice true.


nfagerlund:
:+1:

PR #434, opened 2015-01-27
(maint) Clarify POSTing a delta to a node group.

aperiodic:
The documentation for POSTing to a node group was misleading about the
behavior of the `rule` key. Clarify that so users don't think it's
merged.


ahpook:
:+1: much clearer

PR #437, opened 2015-01-30
Add warning about the PE Infrastructure group

LarissaLane:
DOC-1564

### DIFF
--- a/source/pe/3.7/console_classes_groups_preconfigured_groups.md
+++ b/source/pe/3.7/console_classes_groups_preconfigured_groups.md
@@ -44,11 +44,11 @@ This node group is the parent to all of the node groups listed below (all precon
 
 > **WARNING:** NEVER REMOVE THIS GROUP. 
 > 
-> Removal of the PE Infrastructure node group will bring down all of your PE Infrastructure nodes (Master, PuppetDB, and Console).
+> Removal of the PE Infrastructure node group will disrupt communication between all of your PE Infrastructure nodes (Master, PuppetDB, and Console).
 > 
-> It is very important to correctly configure the `puppet_enterprise` class in this node group. The parameters set in this class affect the behavior of all other preconfigured node groups on this page that use the `puppet_enterprise::profile` class. Incorrect configuration of this class could potentially cause a PE service outage.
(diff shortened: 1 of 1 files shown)


source/pe/3.7/console_classes_groups_preconfigured_groups.md
```diff
@@ -42,6 +42,14 @@ All nodes.
...
 
+> **WARNING:** NEVER REMOVE THIS GROUP. 
+> 
+> Removal of the PE Infrastructure node group will bring down all of your PE Infrastructure nodes (Master, PuppetDB, and Console).
+> 
+> It is very important to correctly configure the `puppet_enterprise` class in this node group. The parameters set in this class affect the behavior of all other preconfigured node groups on this page that use the `puppet_enterprise::profile` class. Incorrect configuration of this class could potentially cause a PE service outage.
+>
+> If you are upgrading from PE 3.3, or need to set up these preconfigured node groups for any other reason, you must configure the PE Infrastructure node group first before any other of the preconfigured node groups (except for the PE MCollective or Default node groups). Failure to do so will cause a disconnection between the Masters, Database, and Console services, and your entire infrastructure will become non-functional.
+
```
Sharpie: @LarissaLane Made a couple tweaks clarifying the type of outage and which classes are dependent on `puppet_enterprise`. How does the following look:

> **WARNING:** NEVER REMOVE THIS GROUP.
>
> Removal of the PE Infrastructure node group will disrupt communication between all of your PE Infrastructure nodes (Master, PuppetDB, and Console).
>
> It is very important to correctly configure the `puppet_enterprise` class in this node group. The parameters set in this class affect the behavior of all other preconfigured node groups on this page that use classes starting with `puppet_enterprise::profile`. Incorrect configuration of this class could potentially cause a PE service outage.
>
> If you are upgrading from PE 3.3, or need to set up these preconfigured node groups for any other reason, you must configure the PE Infrastructure node group first before any other of the preconfigured node groups (except for the PE MCollective or Default node groups). Failure to do so will cause a disconnection between the Masters, Database, and Console services and manual intervention will be required to restore PE Infrastructure functionality.

PR #441, opened 2015-02-02
Style guide

ghoneycutt:


ghoneycutt:
@laurenrother some small things I noticed while reading the new style guide.

source/guides/style_guide.markdown
```diff
@@ -137,7 +137,7 @@ Every publicly available module must have metadata defined in the metadata.json
 ~~~
     {
       "name": "examplecorp-mymodule",
-      "version": "0.0.1",
+      "version": "0.1.0",
```
ghoneycutt: This is here because versioning starts at 0.1.0, so 0.0.1 is not possible with semver.

source/guides/style_guide.markdown
```diff
@@ -163,7 +163,7 @@ Every publicly available module must have metadata defined in the metadata.json
       ],
       "dependencies": [
         { "name": "puppetlabs/stdlib", "version_requirement": ">= 3.2.0 <5.0.0" },
-        { "name": "puppetlabs/firewall", "version_requirement": ">= 0.0.4 <5.0.0" },
```
ghoneycutt: same as above.

laurenrother:
Great catches, @ghoneycutt ! Thanks so much for looking it over :)

PR #442, opened 2015-02-03
Remove smart quotes from Puppet Style Guide

elyscape:
Jekyll/kramdown can add these to the non-code sections and they break the code examples.


laurenrother:
Uuuuuuuugh. Well, that's embarrassing. I thought I'd run a check for those. Apparently not. Thanks so much for the catch/fix!

PR #443, opened 2015-02-03
Make 10.2 example comply with own rules and 10.7

elyscape:
The example in 10.2 violated 10.7 by specifying parameter defaults inside the class. Additionally, it did not adequately demonstrate 10.2.4 or 10.2.8.


laurenrother:
Finally got around to looking this over. Thanks @elyscape !

PR #446, opened 2015-02-05
(DOCUMENT-255) The use of data_binding_terminus to disable automatic hie...

llowder:
...ra lookup should be documented

This patch adds a note to the hiera documentation on how to disable automatic parameter lookups.


nfagerlund:
:+1:

PR #449, opened 2015-02-05
(DOCUMENT-264) Connecting PE to external directory has inconsistent exam...

llowder:
...ple

This patch corrects the examples to make the consistent in terms of the data used.


nfagerlund:
Seems legit. @MFredette, since this touches PE docs, could you merge this and make sure it gets cherry-picked to any work in progress? (And resolve the jira ticket.)

PR #450, opened 2015-02-10
Review and fix future parser doc 3_7

hlindberg:
This provides an update of the future_parser documents to include new language features and remove restrictions that are no longer true.

It also in a few places add TODO notes for suggested changes, or for missing content where it was not clear exactly where such documentation should go.


nfagerlund:
I've gone through all of the changes in this PR and either brought them over, created tickets in the DOC-722 epic to show the need for further work, or rejected them. The results are in commits on the future-parser branch. 

Thanks so much for this feedback Henrik; it took a while to get through, but it was ultimately really helpful.

hlindberg:
You are welcome. Looking forward to reviewing the result !

PR #452, opened 2015-02-10
PuppetDB: Fix command wire format indexes

kbarber:
Previously we had not marged v3 of the reports command format is deprecated
in the index. It was still being pointed at by the index as 'stable' and v4
was only listed in the 'command API' section of the docs.

This corrects that for both 2.2 and master releases.

Signed-off-by: Ken Barber <ken@bob.sh>


nfagerlund:
Cool, looks good to me.

PR #453, opened 2015-02-14
file mode: there is no 4 digit standard

mat-mo:
The file mode 01775 or 02775 are totally acceptable by puppet,
and without the leading 0 one might confuse with decimal. Remove the confusing part.

See also https://github.com/rodjek/puppet-lint/issues/394

### DIFF
--- a/source/references/3.7.4/type.markdown
+++ b/source/references/3.7.4/type.markdown
@@ -1396,9 +1396,8 @@ differently, including Windows. For detailed ACL controls on Windows,
 you can leave `mode` unmanaged and use
 [the puppetlabs/acl module.](https://forge.puppetlabs.com/puppetlabs/acl)
 
-Numeric modes should use the standard four-digit octal notation of
-`<setuid/setgid/sticky><owner><group><other>` (e.g. 0644). Each of the
-"owner," "group," and "other" digits should be a sum of the
+Numeric modes should use octal notation of `<setuid/setgid/sticky><owner><group><other>` (e.g. 0644). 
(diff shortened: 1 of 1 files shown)


Aethylred:
There is a 4 digit recommendation in the Style Guide: https://docs.puppetlabs.com/guides/style_guide.html#file-modes

mat-mo:
I am aware of that, and this PR recommends removing the reference to "4 digit standard" which doesn't exist. Moreover, this specific example is not the best one around.

nfagerlund:
Bunch of fuzzily-overlapping issues here.

* You're right that modes can also be five digits, but the more common deviation is to use three.
* If setuid/setgid/sticky digit isn't specified, puppet CLEARS all those perms. Thus, using three digits is bad because it *looks* less explicit. That's a higher-than-normal priority style issue, worth leaving a note in the type reference. Colloquially, we've been calling that "four digit" because it's so rare to manage su/sg/st. You're right that that's technically wrong, but it gives good results for most users who don't manage etc. etc. etc.
* I have no opinion about leading zeros when there's a nonzero su/sg/st value. That's for the style guide process.
* Here's a PR I just filed that adds more context about the setuid/setgid/sticky bit and says "at least four digits," which I think addresses your concerns. https://github.com/puppetlabs/puppet/pull/3619/files That edit will first appear in the puppet 4.0 docs, which are coming relatively soon.
* I'm closing this PR, because I don't want to delete the "four digit" recommendation without adding more context about setuid/setgid/sticky. Leaving it there will have better results than just deleting it.
* The docs will periodically use "standard" to mean "customary" or "normal."

PR #456, opened 2015-02-21
Guts: Remove all code related to generating PDFs

nfagerlund:


nfagerlund:
@pdxmph Sooooo what do you think.

There was some interesting thinking involved in this, but it was implemented in a messy and hard-to-re-use way, because I had to drop everything and do this in two days. And we can always resurrect it through git history if we need to learn from the past.

nfagerlund:
YANKIN' IT

PR #458, opened 2015-02-26
List of forbidden tags and metadata quality warning.

pizzaops:
Looks like some whitespace also got purged but nothing looks like it should change the formatting...


laurenrother:
Thanks, Zachary!

PR #459, opened 2015-02-27
Move PE to the top of the docs page

ahpook:
Sorry if this ruffles any feathers, but it oughtta be this way. 

PE should get top billing on the docs page and the individual FOSS project drop down underneath it.


pdxmph:
No problem merging. Front page redesign in progress will address this better in the long run.

PR #460, opened 2015-02-27
Fix missing space in config with Hiera example

Sharpie:
Without a space between the key and the value, a Hiera setting won't be picked
up.


nfagerlund:
Yup.

PR #461, opened 2015-03-01
Update plugins in modules

ffrank:
The documentation for Plugins in Modules is quite old. It still documented some limitations that have long been resolved.

This change is not comprehensive - there may be more to clean up. I focused on a particularly alarming falsehood.


nfagerlund:
Wow, this reminds me of surprises I've found cleaning out the fridge. Thanks for catching this!

PR #462, opened 2015-03-02
(maint) update puppetdb catalog versions

wkalt:
Prior to this commit there were some references in the master page to since-removed
command docs.


nfagerlund:
@wkalt Looks good, thanks for the update.

PR #468, opened 2015-03-11
DOCUMENT-298

beergeek:
This fix removes some confusion in relation to trusted facts


nfagerlund:
Oh, good catch, thank you! 

Instead of just deleting, I added an up-to-date note about trusted facts to the summary, and fixed the URL to docs about it while I was in there. See 9918f8c.

PR #471, opened 2015-03-19
(PDB-1142) Update for PuppetDB 2.3.0 release

rbrw:


nfagerlund:
In `_config.yml` we've added a reference to a new navigation sidebar for PuppetDB 2.3, but there doesn't seem to be corresponding file in `source/_includes`. Need to add one of those.

nfagerlund:
Closed in favor of #472

PR #476, opened 2015-03-24
Lift the requirement for aligned hashrockets

nibalizer:
Hashrockets have been required to be aligned for a long time. This
has long been attributed to readability. I wonder if there is real
data to support this, or if readability is a matter of preference.

At OpenStack, we find that the aligned hashrockets restriction causes
unnecessary churn in the code base. Every time a new parameter is added
with a different length, the entire code bock shifts in or out.

There are two major problems from this. Removing the restriction removes
the problem:

1) git-blame doesn't show correct information. Instead of showing the
last person/commit that materially changed the line in question, it shows
the last person/commit that tabbed the line in or out.
There are whitespace-ignoring options that can be used on the command
line, but there is a lot of tooling around git and not all of it runs
in this mode.

2) Code review has more code changing. What should be a one line change
to review now looks like several other parameters have been changed or
added. At OpenStack we review a lot of code, increasing the speed that
we can review at is the single biggest way to move our project forward
faster.

This came out of a discussion this review:
https://review.openstack.org/#/c/164819/

### DIFF
--- a/source/guides/style_guide.markdown
+++ b/source/guides/style_guide.markdown
@@ -61,7 +61,6 @@ Module manifests:
 * Must have trailing commas after all resource attributes and parameter definitions,
 * Should not exceed a 140-character line width,
 * Should leave one empty line between resources, except when using dependency chains, and
-* Should align hash rockets (`=>`) within blocks of attributes, remembering to arrange hashes for maximum readability first.
 
 ## 6. Quoting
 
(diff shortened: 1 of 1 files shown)


daenney:
I do find it more readable when things are aligned. My head tends to work in block editing mode which makes a "rigid" block like:

```puppet
file { foo:
  ensure => yolo,
  owner  => 0, 
  mode   => '0644',
  group  => 0,
}
```

Easier to spot and act on than this "wobbly" block:
```puppet
file { foo:
  ensure => yolo,
  owner => 0, 
  mode => '0644',
  group => 0,
}
```

That said I do agree that the extra white space changes often take me extra time during reviews because I need to double check them just to be sure some additional change didn't get sneaked in with them.

laurenrother:
Howdy @nibalizer ! There's going to be another big release of the style guide for compatibility with Puppet 4. We're tracking all suggestions, ideas, and opinions here: https://tickets.puppetlabs.com/browse/DOCUMENT-266 Totally feel free to just copy/paste this into a ticket attached to that one and we'll go over it with the whole team :)

nibalizer:
https://tickets.puppetlabs.com/browse/DOCUMENT-306 created, thanks :)

PR #482, opened 2015-03-27
Fix typo

juniorsysadmin:


nfagerlund:
This version of the PuppetDB docs is dead and frozen, and we're not accepting updates to it. The current versions of the PuppetDB docs live in the puppetlabs/puppetdb repo, in the `documentation` folder. Thanks for being on the lookout, though!

PR #483, opened 2015-04-02
DOCUMENT-312: change API-link to use framed version of YARD output

DavidS:


nfagerlund:
Yeah, good call, I agree with doing this. 

Could you please amend this commit to leave off the `#!file.README.html` anchor part on each link?

DavidS:
Done.

nfagerlund:
Grand, thank you!

PR #484, opened 2015-04-03
Add missing links for local scope to data type documentation

elyscape:
The last paragraph of `lang_datatypes.markdown` for all versions since
2.7 has a reference to local scopes that is marked as a link, but no URL
to accompany it.


nfagerlund:
Oh yup, good catch.

PR #490, opened 2015-04-15
(maint) Add missing image for installing MSI with GUI

joshcooper:
This image is different from the other versions in that this one doesn't
reference Puppet Enterprise, and it refers to Puppet Agent (64-bit).


ferventcoder:
+!

PR #494, opened 2015-04-17
(DOC-1887) Puppet 4 URL changes

rlinehan:
Update URLs in Puppet docs to conform to Puppet 4 URL changes. This has almost no content changes, I just went through and changed all the URLs I could find.

The first 2 commits just copy files - guides that are Puppet 4 specific I just copied to files that end with `-puppet-4`, and references I copied over from puppet/3.7 to puppet/4.0. Te rest of the commits show the actual changes I made.


nfagerlund:
I rebased and merged the reference commits into the puppet4.1 branch, in merge commit 2885a59; the guides changes, I'm leaving behind because we plan to basically dynamite those pages. 

Thanks again for getting these!

PR #506, opened 2015-05-19
Maybe give a little deeper explanation about puppet server

ahpook:


nfagerlund:
Exactly what we needed. Thanks!

PR #507, opened 2015-05-21
platforms-update

kylog:
@nfagerlund I'm not sure if this page is getting more drastic revamps in the near future or not. If it is, disregard/close this. If it's not, I can tack on another commit adding puppet 4.x and ruby 2.2.


nfagerlund:
Hahaha, crikey! I haven't looked at this page in forever. Thanks, I'm merging this update, but we might want to blow this away in favor of version-specific system requirements sometime soon.

PR #509, opened 2015-05-28
platforms-update-p4

kylog:


kylog:
@nfagerlund this is a follow-up to #507. I tweaked wording from 'Supported' to 'Tested' - let me know what your thoughts are on that (and the other changes).

nfagerlund:
cool cool. I'm okay with that; went ahead and merged this after doing a few minor formatting tweaks.

PR #512, opened 2015-06-09
(DOCUMENT-368) alternate ways to get 64 bit paths

ferventcoder:
On Windows, the search path is able to be used to search for a process. This allows
us to take advantage of using the search path to enter a specific order of searches
for an executable so that we are always using the 64-bit version or alternatively,
always selecting the 32-bit version based on where the executable is found first on
the path. This allows a process that is sometimes in sysnative and sometimes in
system32 based on whether we are in a 32-bit process on a 64-bit OS or in a 32-bit
process on a 32-bit OS. Either way the proper item will be selected provided the
search strings are correct.

### DIFF
--- a/source/windows/troubleshooting.markdown
+++ b/source/windows/troubleshooting.markdown
@@ -122,6 +122,8 @@ Instead, wrap the builtin in `cmd.exe`:
       path => 'c:\windows\system32;c:\windows'
     }
 
+Keep in mind that in 32-bit versions of Puppet, you may be subject to file system redirection (where system32 is switched to sysWOW64 automatically by Windows). We have ways of dealing with the [redirection](/puppet/latest/reference/lang_windows_file_paths.html#file-system-redirection-when-running-32-bit-puppet-on-64-bit-windows).
+
 Or, better still, use the tip from above:
 
     exec { 'cmd.exe /c echo foo':
--- a/source/puppet/3.7/reference/future_lang_windows_file_paths.markdown
+++ b/source/puppet/3.7/reference/future_lang_windows_file_paths.markdown
@@ -156,6 +156,20 @@ Prior to 3.7.3, you can manually compensate. On systems affected by file system 
 
 Prior to 3.7.3, there's no easy way for Puppet manifests to detect whether `sysnative` is available or necessary. Authors of public modules can choose to only support 3.7.3+, or can ship a renamed version of the `$system32` fact to support older versions. Private users can predict the mix of OS versions and architectures where their code will be run, and simply do the right thing for that environment.
 
+One consideration for `exec` when you can't use `$system32` is to use `path =>` and set it appropriately, the first search path item first followed by others in the order you want them searched in. For instance, if you want to always use the 64-bit version of `cmd.exe` you can use:
+
+    exec { '64_bit_cmd':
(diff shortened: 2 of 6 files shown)


ferventcoder:
This does need to be tested to prove that the output is correct. I did cursory checking to be sure the output looked mostly correct, but didn't run the full site.

nfagerlund:
By "output is correct," you just mean the markdown formatting, right? Looks good to me. 

Three things: 

* Is that example meant to use the $path fact, rather than pass the string $path to windows? If so, maybe use $::path to show the reader what we mean. 
* This hint only applies to `exec` resources, so we should say so before going into details. The rest of the page applies to `file`, `exec`, custom plugins, and others. 
* There are two more places to paste this in: the 3.7 and 3.8 docs have a "future_lang_windows_....." file. 

Once those are addressed, this looks good.

ferventcoder:
Future lang - sweet, I would have missed those.

ferventcoder:
> Is that example meant to use the `$path` fact, rather than pass the string $path to windows? If so, maybe use `$::path` to show the reader what we mean.

I thought we were getting away from top-level name-spacing facts, that's why I didn't. But I can see why there may be confusion. On windows though, it is `%PATH%` or `$env:PATH`. I will make the adjustment though.

source/windows/troubleshooting.markdown
```diff
@@ -122,6 +122,8 @@ Instead, wrap the builtin in `cmd.exe`:
       path => 'c:\windows\system32;c:\windows'
     }
 
+Keep in mind that in 32-bit versions of Puppet, you may be subject to file system redirection (where system32 is switched to sysWOW64 automatically by Windows). We have ways of dealing with the [redirection](/reference/lang_windows_file_paths.html#file-system-redirection-when-running-32-bit-puppet-on-64-bit-windows).
+
```
ferventcoder: This link may need to be validated. I wanted to point that out from troubleshooting how to deal with file system redirection.

ferventcoder:
@nfagerlund all issues addressed.

source/windows/troubleshooting.markdown
nfagerlund: It's not quite right. You'll want `/puppet/latest/reference/lang_windows_file_paths.html#file-system-redirection-when-running-32-bit-puppet-on-64-bit-windows`

nfagerlund:
Yeah, the thing about $:: is that it's almost never _necessary_ to retrieve the right value, but it's helpful when you intend the code to be read. 

The real solution is moving to `$facts[path]` in all our examples once Puppet 4 is well and truly entrenched, but that'll be a minute.

source/windows/troubleshooting.markdown
ferventcoder: Right on. I went off of another example, so let me clean that up.

ferventcoder: Fixed.

ferventcoder:
Okay ready ready haha

nfagerlund:
:+1:

nfagerlund:
@ferventcoder Thanks man.

ferventcoder:
Will this go up somewhere for review soon?

nfagerlund:
Should be live already: https://docs.puppetlabs.com/puppet/latest/reference/lang_windows_file_paths.html#compensating-for-redirection

PR #514, opened 2015-06-16
(maint) Windows enhancements

ferventcoder:
* With MSIExec, always add `/norestart` to the options.
* In the deploy non-root agent guide, be sure to add extra information so that users know we really don't recommend that setting with Windows. It's a path of trouble and provides an extremely limited use of Puppet - Windows just doesn't have the same non-root access that *nix provides.


ferventcoder:
@nfagerlund For your review sir.

nfagerlund:
Sorry this took a while! OK, let's get it done.

ferventcoder:
Thanks!

PR #515, opened 2015-06-18
fix a typo in 4.x reference/whered_it_go.markdown

kpankonen:


nfagerlund:
Good catch, thank you.

PR #520, opened 2015-07-02
Facter 3.0.0 release notes re: FACT-1085

gguillotte:
Added sections noting the regression reported by FACT-1085 (Facter 3 incorrectly redirects stderr to stdout) and stderr break also noted in FACT-1085 (Facter 3 no longer passes stderr, use --debug instead).


gguillotte:
Updated per Nick's suggestions.

PR #522, opened 2015-07-08
Release PuppetDB 3.0

nfagerlund:
Unlock latest, move it to 'current versions' in the index, and change version notes in _config.yml.


nfagerlund:
This is to be merged by the PuppetDB team when they're ready to ship.

PR #524, opened 2015-07-14
DOCUMENT-385: Clarify exportable resources.

gguillotte:
In all articles that use text from `lang_exported` (including future
parser documentation and the Exported Resources guide), add a note
clarifying that only node-specific resources publishable by the catalog
are exportable, and arbitrary data not stored or published by the
catalog are not.

### DIFF
--- a/source/puppet/4.2/reference/lang_exported.markdown
+++ b/source/puppet/4.2/reference/lang_exported.markdown
@@ -32,9 +32,9 @@ An **exported resource declaration** specifies a desired state for a resource, *
 Purpose
 -----
 
-Exported resources allow nodes to share information with each other. This is useful when one node has information that another node needs in order to manage a resource --- the node with the information can construct and publish the resource, and the node managing the resource can collect it.
+Exported resources allow the Puppet compiler to share information among nodes by combining information from multiple nodes' catalogs. This helps you manage things that rely on nodes knowing the states or activity of other nodes.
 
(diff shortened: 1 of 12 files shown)


source/puppet/4.2/reference/lang_exported.markdown
```diff
@@ -34,6 +34,8 @@ Purpose
 
 Exported resources allow nodes to share information with each other. This is useful when one node has information that another node needs in order to manage a resource --- the node with the information can construct and publish the resource, and the node managing the resource can collect it.
```
nfagerlund: I think this graf wants some revision as well. The key thing I did wrong here was saying they "allow nodes to share information with each other" — it's not _quite_ the nodes doing the sharing, it's the compiler... but it's hard to talk about that. 

Maybe we should say something like,

> Exported resources let you combine information from multiple nodes' catalogs. This is useful when you're managing something that depends on what other nodes are doing. 

?

source/puppet/4.2/reference/lang_exported.markdown
```diff
@@ -34,6 +34,8 @@ Purpose
...
 
+**Note**: Exported resources only include node-specific data, and the catalog compiler constructs and publishes exported resources from that data. This means you can only export node-specific resources stored in the catalog. Generic data not stored in the catalog, such as an entire arbitrary file on a node's filesystem, can't be exported even if referenced by the source parameter of a file resource.
```
nfagerlund: This could stand to be smaller and tighter. I think the way to accomplish that is to not talk about node-specific `_____` and focus on the compiler / agent boundary. Something like, "Exported resources can only use the information that's available to the compiler; they can't use information that was never sent to the compiler, like the contents of an arbitrary file on some node."

PR #527, opened 2015-07-17
Updated caveat about supplying stage to a class resource-style

pizzaops:
Stages not only cannot be set for a class while using the `include` function, they can't be set via hiera at all, even when declaring the class resource-style.

I also removed the duplicate description of that caveat, since they were already right next to each other in the doc anyway.


nfagerlund:
Cool, thanks.

PR #529, opened 2015-07-23
Clarify upgrade order

MikaelSmith:
The server upgrade docs point out upgrading the server prior to
upgrading the agent; that point seems like it should be made during the
preparation phase. Change the order of presenting agent and server
upgrade docs - server should really be addressed first - and explicitly
point out this ordering.

### DIFF
--- a/source/puppet/4.2/reference/upgrade_major_pre.md
+++ b/source/puppet/4.2/reference/upgrade_major_pre.md
@@ -80,4 +80,4 @@ Puppet 4.0 introduces several breaking changes, some of which didn't go through 
 
 ## You're Ready!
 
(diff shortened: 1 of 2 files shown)


MikaelSmith:
@gguillotte It seemed like something should specifically be said about upgrading the server before agents in the pre-upgrade docs, so I put something together.

nfagerlund:
I'm in the process of reviewing these, so I'll merge.

MikaelSmith:
Ok; I meant it more as a conversation-starter, but I assume you'll refine the language as needed.

nfagerlund:
yup yup

PR #530, opened 2015-08-05
DOC-2188: Adjust Puppet Server branches per SERVER-828

nfagerlund:
DO NOT MERGE UNTIL SERVER-828 IS RESOLVED.

2.1.x is moving to stable, and 1.1.x is moving to 1.x. I'm also re-enabling
master, so there's a way to preview what the docs look like for the next
version.


nwolfe:
:+1: Branches have moved! We're ready for this now.

PR #534, opened 2015-08-13
DOC-2203: Document new reserved words.

gguillotte:
Document `application`, `consumes`, and `produces` in the 3.8 and 4.2
references. Note in the 3.8 reference that these words are only reserved in
the future parser.

### DIFF
--- a/source/puppet/3.8/reference/lang_reserved.markdown
+++ b/source/puppet/3.8/reference/lang_reserved.markdown
@@ -31,7 +31,7 @@ Several words in the Puppet language are **reserved**. This means they:
 * Cannot be used as names for classes.
 * Cannot be used as names for custom resource types or defined resource types.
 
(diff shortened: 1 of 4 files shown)


source/puppet/3.8/reference/lang_reserved.markdown
```diff
@@ -60,6 +60,14 @@ The following words are reserved for future use:
 * `type`
 * `private`
 
+The following words are reserved when using the future parser:
```
nfagerlund: There's actually a separate file you'll need to update as well: future_lang_reserved. There are two copies of the language docs in 3.8, since there are two copies of the language.

nfagerlund:
@gguillotte Other than updating the future_lang_reserved page, I think this is good to go.

gguillotte:
Updated, pushed, and confirmed changes.

nfagerlund:
GUH, I forgot about this PR, and made an almost identical commit in the Puppet 4.2.2 branch. 

I'm going to close this and let the one in 4.2.2 merge to master when we ship that release. SORRY :( :( :(

PR #540, opened 2015-09-14
Update path in AWS trusted facts example.

gguillotte:


source/puppet/4.2/reference/ssl_attributes_extensions.markdown
```diff
@@ -171,53 +174,51 @@ Puppet Labs provides several registered OIDs (under "ppRegCertExt") for the most
...
-### Recovering From Failed Data Embedding
+When first testing this feature, you might fail to embed the right information in a CSR or certificate and want to start over for your test nodes. (This is not really a problem once your provisioning system is changed to populate the data, but it can happen pretty easily when doing things manually.)
 
-When first testing this feature, you might fail to embed the right information in a CSR or certificate and want to start over for your test node(s). (This is not really a problem once your provisioning system is changed to populate the data, but it can happen pretty easily when doing things manually.)
+To start over, do the following:
 
-In order to start over, you'll need to do the following:
-
-**On the test node:**
+**On the test node**:
```
nfagerlund: Not going to block on it, but this is typographically wrong; punctuation adjacent to text should have the same font styling as the text it adjoins.

PR #544, opened 2015-09-25
Update facter docs to replace `facter` with `puppet facts`

abottchen:
For debugging the situation where external facts are not seen by puppet,
the documentation tells customers to look at `facter --debug`.  This
output will not contain external facts.  The correct command for this
purpose would be `puppet facts --debug`.


nfagerlund:
Seems sensible enough, sure.

PR #545, opened 2015-10-06
(DOCUMENT-436) Puppet report format changes tags from an array

fiddyspence:
Puppet report format changes tags from an array to a Puppet::Util::TagSet object


nfagerlund:
I want to wait to merge this until we get feedback from the client team on PUP-5341.

nfagerlund:
Went with a different fix in 704108c.

PR #547, opened 2015-10-07
DOCUMENT-437: Correct external facts directories for Facter 3.x

jearls:
This PR adds the `/opt/puppetlabs/facter/facts.d` directory to the Unix/Linux/MacOS X list of external facts directories.


nfagerlund:
Legit. Thanks for catching this!

PR #551, opened 2015-10-14
Emphasize unstructured fact deprecation.

gguillotte:
Facter's core facts list notes that legacy unstructured facts are hidden from command-line output in favor of structured facts, but doesn't emphasize that unstructured facts are deprecated. Emphasize this, both in the summary at the top of the list and in the individual legacy facts.

This doc is a good candidate for a more ambitious reorganization as a followup to segment this list into structured and unstructured facts.

### DIFF
--- a/source/facter/3.0/core_facts.md
+++ b/source/facter/3.0/core_facts.md
@@ -4,27 +4,25 @@ title: "Facter 3.0: Core Facts"
 toc: columns
 ---
 
+This is a list of all of the built-in facts that ship with Facter 3.0, which includes both legacy facts and newer structured facts.
 
-This is a list of all of the built-in facts that ship with Facter 3.0.
+Not all of them apply to every system, and your site might also use [custom facts](./custom_facts.html) delivered via Puppet modules. To see the full list of structured facts and values on a given system (including plugin facts), run `puppet facts` at the command line. If you are using Puppet Enterprise, you can view all of the facts for any node on the node's page in the console.
 
-Not all of them apply to every system, and your site might also use [custom facts](./custom_facts.html) delivered via Puppet modules. To see the full list of facts and values on a given system (including plugin facts), run `puppet facts` at the command line. If you are using Puppet Enterprise, you can view all of the facts for any node on the node's page in the console.
+You can access facts in your Puppet manifests as `$fact_name` or `$facts[fact_name]`. For more information, see [the Puppet docs on facts and built-in variables.](/puppet/latest/reference/lang_facts_and_builtin_vars.html)
 
-You can access facts in your Puppet manifests as `$fact_name` or `$facts[fact_name]`. For more info, see [the Puppet docs on facts and built-in variables.](/puppet/latest/reference/lang_facts_and_builtin_vars.html)
-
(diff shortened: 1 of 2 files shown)


gguillotte:
Hot off the presses, @kylog @jessereynolds !

jessereynolds:
:+1: nice one, thanks @gguillotte

source/facter/3.0/core_facts.md
```diff
@@ -4,20 +4,19 @@ title: "Facter 3.0: Core Facts"
 toc: columns
 ---
 
-
 This is a list of all of the built-in facts that ship with Facter 3.0.
```
kylog: Since this is the introductory line of text, maybe change this line to call out the distinction from the get-go. E.g. something like:

"Above is a list of all the built-in facts that ship with Facter 3.0. Please note that built-in facts include both legacy facts (which are deprecated) and newer structured facts."

Note also that I used the term "legacy" above rather than "unstructured." I'm not totally wedded to that, but here's why I phrased it that way (and btw: pedant alert): there are one or two built-in facts which are both legacy and structured, i.e. because they were introduced in facter 2 (when we first added structured facts) but then restructured in facter 3 (when we structured *everything* and thought the structure through systematically).

In short, "legacy" is a bit more accurate than "unstructured" and may also connote "stop using this" more. It also matches the `--show-legacy` command line option. But I'm terrible at naming things :)

If we do opt for "legacy" here, let's make that change in the per-fact lines modified below as well.

source/facter/3.0/core_facts.md
```diff
@@ -4,20 +4,19 @@ title: "Facter 3.0: Core Facts"
...
 
-Note that some facts, like `architecture`, are hidden in Facter's command-line output. These are all older facts that have become part of a more useful structured fact; for example, `architecture` is now part of the `os` fact. You can still use the old versions in Puppet manifests (e.g. `$architecture`), and you can still specifically request them on the command line (e.g. `facter architecture`), but we suppress them to make the default output less noisy.
+> **Deprecation Note:** As of Facter 3, legacy unstructured facts such as `architecture` are deprecated and hidden to reduce noise in Facter's default command-line output. These older facts are now part of more useful structured facts; for example, `architecture` is part of the `os` fact, accessible as `os.architecture`. You can still use these deprecated unstructured facts in Puppet manifests (`$architecture`), request them on the command line (`facter architecture`), and view them alongside structured facts (`facter --show-legacy`).
 
 * * *
 
 ## `architecture`
 
-This fact is hidden in Facter's command-line output.
+This unstructured fact is deprecated and hidden in Facter's command-line output.
```
kylog: Should we replace "hidden" with "hidden by default" in all these?

kylog:
This looks awesome @gguillotte! Thanks for jumping right on this. I had a couple minor ideas for tweaks but I'm :+1: regardless :)

nfagerlund:
@gguillotte, what do you think about @kylog's suggestion re: "legacy?"  I think I'm pretty much in favor of it, but we'll also have a chance to revisit this when we modify the template to split these into different pages.

gguillotte:
I don't have any issues with legacy, but I think we briefly touched offline (and also back when Facter 3 came out, in discussions about `--show-legacy` in the release notes) about the implications of "deprecated" when referring to these facts. Have they officially been deprecated? It was up in the air at the time of the Facter 3 release notes.

gguillotte:
Pushed a new version that takes the above into consideration. I've taken "deprecated" out but can easily put it back in where appropriate if those facts are officially deprecated.

PR #553, opened 2015-10-15
Add note to Agent release page explaining 1.2.6.

gguillotte:


gguillotte:
@ahpook @kayayarai

ahpook:
Hm, I think I'd rather see a section header at the same level as "Release Contents", named "Release Notes", that describes each of the version bumps.  It doesn't have to go back in time very far, maybe 1.2.0 onward, and should just pull the headline from the puppet-announce email.

* puppet-agent 1.2.6: (the thing you wrote)
* puppet-agent 1.2.5: 
  * Includes mcollective 2.8.6, which fixes an issue when trying to start 
   mcollectived on Solaris 10. 
  * Changes the package filenames on Mac OS X to use major and minor versions, 
   e.g. puppet-agent-1.2.5-1.osx10.10.dmg, instead of codenames, e.g. 
   puppet-agent-1.2.5-1.yosemite.dmg 
* puppet-agent 1.2.4: Component updates including new Puppet, Facter and Mcollective releases
* puppet-agent 1.2.3: Was not released, as regressions were discovered after tagging
* puppet-agent 1.2.2: Bugfix releases in Facter and Puppet
* puppet-agent 1.2.1: Not released(?)
* puppet-agent 1.2.0: Included Puppet 4.2 and Facter 3, the C++-based Facter rewrite.
* For details on earlier puppet-agent versions, please see the relevant post in the [puppet-announce google group](https://groups.google.com/forum/#!forum/puppet-announce)

That way it's less of a one-off and shows a trackable history of the agent itself.  @nfagerlund may have longer-term plans for where to put this info in the docs IA, but in my opinion it's important to get this up inline now and move it later because there are quite a number of requests and no other reasonable place for people to find this information.

nfagerlund:
Garrett and I just talked about this, and I'd like to put an include in the source/puppet-agent folder, like we do with the tables. Then we can pull the include in on that about_agent page. 

I'd be in favor of including those short notes about prior versions, but that can depend on time.

nfagerlund:
so mote it be

PR #554, opened 2015-10-17
(PUP-5395) allow deep merging on automatic parameter lookup

rnelson0:
Document the two steps required to perform deep merges via APL.

### DIFF
--- a/source/hiera/3.0/lookup_types.markdown
+++ b/source/hiera/3.0/lookup_types.markdown
@@ -277,6 +277,26 @@ With a `deep` merge, you would get:
 
 In this case, deglitch.yaml was able to set the group because common.yaml didn't have a value for it, but where there was a conflict, like the uid, common won. Most users don't want this.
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
@rnelson0, thanks for this docs proposal! I really appreciate the early warning. 

Rest of docs team: this can't be merged, because it describes a proposed feature that isn't part of any planned Puppet version yet. It sounds like there's widespread agreement about having the lookup system control its data, but there are unanswered questions about how this would interact with the 4.3 Puppet lookup work, and how to implement this in a forward-compatible way. 

I should probably close the PR until we have a decision and then reopen it, but for now I'm going to leave it open; as 4.3 gets closer, I'll close the PR if it isn't going into that release, and we can dig it back up later.

rnelson0:
PUP-5395 is now being worked by the Puppet Language team to get it into 4.3.0, but I am glad to update this when the solution is finalized.

rnelson0:
Closing this out, PUP-5395 has progressed quite a bit and changed the setup. I'll open a new PR when it's ready. Thanks.

hlindberg:
The work on PR-5395 is now merged

PR #559, opened 2015-10-28
Update docs on refreshing/notification in Puppet language

nfagerlund:
I realized we were missing some detailed info about how this actually works.

I could use a second set of eyes on this, to make sure it makes sense.


nfagerlund:
Got a :+1: from @kayayarai

PR #561, opened 2015-10-28
Add puppet-agent 1.2.7 release notes.

gguillotte:
Adds release notes.

### DIFF
--- a/source/puppet-agent/_release_notes.md
+++ b/source/puppet-agent/_release_notes.md
@@ -1,10 +1,13 @@
 ### Release Notes
 
+* **1.2.7**, Oct. 29, 2015: Updates Ruby on Windows, [Puppet](/puppet/4.2/reference/release_notes.html#puppet-423), [Facter](/facter/3.1/release_notes.html#facter-311), and [Hiera](/hiera/latest/release_notes.html#hiera-304). This version also [moves the OS X bill of materials](https://tickets.puppetlabs.com/browse/PA-21) to `/usr/local/share` on all supported OS X platforms to ensure future compatibility with OS X 10.11 (El Capitan), [adds a root certificate](https://tickets.puppetlabs.com/browse/PA-20) in Enterprise Linux 4 to enable Forge access, and [resolves a security issue](https://tickets.puppetlabs.com/browse/PA-27) in Windows.
+
+    Puppet Enterprise 2015.2.3 will also offer packages for AIX and Solaris 11.
 * **1.2.6**, Oct. 12, 2015: Contains no component fixes or updates. It only resolves a packaging issue that could leave the `puppet` and `mcollective` services stopped and unregistered after a package upgrade from an earlier `puppet-agent` release.
 * **1.2.5**, Oct. 1, 2015: Updates [MCollective](/mcollective/releasenotes.html#changes-since-284) to fix an issue when trying to start `mcollectived` on Solaris 10. It also changes the package file names on OS X to use major and minor OS versions (such as `puppet-agent-1.2.5-1.osx10.10.dmg`) instead of codenames (such as `puppet-agent-1.2.5-1.yosemite.dmg`).
 * **1.2.4**, Sept. 14, 2015: Updates [Puppet](/puppet/4.2/reference/release_notes.html#puppet-422), [Facter](/facter/3.1/release_notes.html#facter-310), and [MCollective](/mcollective/releasenotes.html#changes-since-283).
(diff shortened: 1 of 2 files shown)


gguillotte:
@joshcooper @kylog @nfagerlund

source/puppet-agent/_release_notes.md
```diff
@@ -1,10 +1,13 @@
 ### Release Notes
 
+* **1.2.7**, Oct. 29, 2015: Adds or improves support for AIX 5.3, [6.1](https://tickets.puppetlabs.com/browse/PA-32), and [7.1](https://tickets.puppetlabs.com/browse/PA-33) on PowerPC, Solaris 11 on [SPARC](https://tickets.puppetlabs.com/browse/PA-4) and [i386](https://tickets.puppetlabs.com/browse/PA-5), and Enterprise Linux 4. Updates Ruby, [Facter](/facter/3.1/release_notes.html#facter-311), and [Hiera](/hiera/latest/release_notes.html#hiera-304).
```
joshcooper: Could you link AIX 5.3 to https://tickets.puppetlabs.com/browse/PA-41 (just moved the RE ticket to PA like the others)

joshcooper:
@gguillotte I just remembered that the new platform additions (EL4, AIX, Solaris 11) are PE only. This means we will not be providing open source packages for those platforms when we ship puppet-agent 1.2.7, but they will be available for download for PE customers when PE 2015.2.3 comes out in Nov.

Since we're not shipping any bits for those platforms at this time, I'm thinking we should remove references to those platforms, and handle it in PE documentation. Thoughts? /cc @ipeldridge @nfagerlund

ipeldridge:
@joshcooper I've already added them to the docs for PE 2015.2.3. Thanks.

kylog:
I agree with removing Sparc/AIX/EL4 from the 1.2.7 release notes.

gguillotte:
Removed. Does this affect the Facter and Hiera release notes as well?

kylog:
For any of the components (e.g. puppet/facter/hiera), I think it's correct/appropriate to mention fixes in the source for any/all platforms, including the packaged-for-PE-only platforms. So my comment above was specific to puppet-agent. (Otoh, once we open-source puppet-agent, we might want to fine-tune this reasoning.)

Re hiera specifically, though, IIRC there are no functional changes, so the hiera release notes would just be like for 3.0.3, no functional change: https://docs.puppetlabs.com/hiera/3.0/release_notes.html.

I know this may be kind of hair-splitting. Do those approaches make sense?

gguillotte:
Yep! Thanks @kylog .

joshcooper:
:+1: to what @kylog said though there was one (maint) commit to hiera that changes an error message when failing to retrieve hiera data. Not sure if we want to mention that for 3.0.4 specifically?

joshcooper:
Here are top-level queries:

introduced in puppet-agent 1.2.7: https://tickets.puppetlabs.com/issues/?filter=15776
fixed in puppet-agent 1.2.7: https://tickets.puppetlabs.com/issues/?filter=15775

gguillotte:
Amended to clarify that PE packages will include AIX and Solaris 11.

source/puppet-agent/_release_notes.md
```diff
@@ -1,10 +1,13 @@
 ### Release Notes
 
+* **1.2.7**, Oct. 29, 2015: Updates Ruby, [Facter](/facter/3.1/release_notes.html#facter-311), and [Hiera](/hiera/latest/release_notes.html#hiera-304). This version also [moves the OS X bill of materials](https://tickets.puppetlabs.com/browse/PA-21) to `/usr/local/share` on all supported OS X platforms to ensure future compatibility with OS X 10.11 (El Capitan), and [adds a root certificate](https://tickets.puppetlabs.com/browse/PA-20) in Enterprise Linux 4 to enable Forge access.
```
joshcooper: I think we only bumped Ruby to 1.0.2d for Windows, as the other platforms were already there.

Also could you add a note that puppet was updated too?

kayayarai:
:+1:

joshcooper:
@gguillotte looks like my comment is getting lost in the force push. Could you add a note that Puppet was updated too (right now it just mentions Facter and Hiera). Also we only updated Ruby to 1.0.2d for Windows, but not for any other platform.

gguillotte:
@joshcooper Bumped.

joshcooper:
:+1:

PR #568, opened 2015-11-05
Clarify that you don't need to restart for environment.conf changes

pizzaops:


nfagerlund:
Hmm. 

* Good idea to say this, since it's such a departure from puppet.conf (which sometimes even has the same settings!). 
* After thinking a bit, I figure this is the wrong place in the page for it, and also I'm thinking: we should probably have a section about reload behavior on EVERY config file page. 

@pizzaops, won't ask you to do any of that. Docs team will handle it. Thanks for the nudge! https://tickets.puppetlabs.com/browse/DOC-2346

PR #571, opened 2015-11-09
Add deprecation notice for implicit file symlinks.

gguillotte:
File resources let you create symlinks through arbitrary string values
for the `ensure` attribute. This behavior is being deprecated; document
the deprecation.

See also puppet [PR 4428](https://github.com/puppetlabs/puppet/pull/4428).


kylog:
:+1:

PR #573, opened 2015-11-11
(docs) Remove references to cfacter

kylog:
We're not shipping cfacter any more, so remove references to it.


kylog:
@gguillotte or @nfagerlund; I just noticed that we have both some sample code referencing `cfacter` and some other misc references. Since we haven't shipped a separate `cfacter` binary since the puppet-agent 1.1.x series, we should probably drop references to it.

Assuming we do so, I don't know if we'd want to roll this back into the 4.2 or earlier directories, or not. 

Thoughts?

gguillotte:
@kylog :+1: Makes sense to me. We can merge this and I can make the same changes to the 4.1 docs. cc: @nfagerlund

PR #574, opened 2015-11-11
Remove references to cfacter from 4.2.

gguillotte:
See also PR #573. cc @kylog @nfagerlund

### DIFF
--- a/source/puppet/4.2/reference/lang_functions.markdown
+++ b/source/puppet/4.2/reference/lang_functions.markdown
@@ -72,7 +72,6 @@ file {"/etc/ntp.conf":
 include apache # function call; modifies catalog
 
 $binaries = [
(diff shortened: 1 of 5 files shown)


gguillotte:
@kylog In the `puppet` repo, there's a reference to the `cfacter` configuration setting in /man/man5/puppet.conf.5. Are the manpages generated from something else, or can we edit them directly?

nfagerlund:
@gguillotte @kylog That man page is generated from the configuration reference, which is generated in turn from lib/puppet/defaults.rb. 

But... I don't know whether there's automation to keep the "frozen" troff-formatted copy in the puppet repo up to date. My suspicion is that there isn't.

nfagerlund:
That aside, I'm gonna go ahead and merge this.

PR #577, opened 2015-11-16
Updating facter facts in hiera documentation with best practice

MrHyatt:
Changed %{fqdn} to %{::fqdn} per best practice

Nothing fancy


nfagerlund:
I squashed these commits into one, and merged it into master as 8cf67b1. Thank you!

PR #579, opened 2015-11-19
Update System Requirements for 4.3.

gguillotte:
We added and removed platforms in `puppet-agent` 1.3.0 and Puppet 4.3.0.
The System Requirements page (and related pages) should reflect this.

This commit also fixes broken image links in the Install Windows page.

Please confirm that the Fedora 22 packages are live before merging.


source/_includes/pup43_platforms_fedora.markdown
```diff
@@ -0,0 +1,7 @@
+We publish official packages and run automated testing on the following versions for Puppet 4.3:
```
melissa: Hm, this is a little misleading. We don't actually publish puppet 4 packages. It's a part of the puppet-agent package. Given that we're trying to help people move over to the "new mentality" of AIO-land, we might want to rethink this wording.

source/_includes/pup40_platforms_redhat_like.markdown
```diff
@@ -1,10 +1,10 @@
 We publish official packages and run automated testing on the following versions for Puppet 4.x:
 
-* Enterprise Linux 6 (also supported by [Puppet Enterprise][peinstall])
 * Enterprise Linux 7 (also supported by [Puppet Enterprise][peinstall])
-
+* Enterprise Linux 6 (also supported by [Puppet Enterprise][peinstall])
+* Enterprise Linux 5 (also supported by [Puppet Enterprise][peinstall])
```
melissa: EL 5 is only a supported as an agent platform. I'm not sure if/where that's already being called out somewhere though

source/_includes/pup43_platforms_fedora.markdown
```diff
@@ -0,0 +1,7 @@
...
+
+* Fedora 22
+* Fedora 21
+* Fedora 20
```
melissa: Fedora 20 is EOL, so we aren't going to build any new packages for it. But since these are puppet 4.3 docs, that might not be an applicable point

source/_includes/pup40_platforms_redhat_like.markdown
melissa: Hmm, wait, we do have el5 server packages (puppetserver and puppetdb). I'm trying to remember if that was a mistake or intentional... I'm fairly sure that was a mistake...

melissa: ah, yeah, definitely :(

https://confluence.puppetlabs.com/display/PROD/FOSS+Supported+Platforms+for+PC1

cc/ @underscorgan

underscorgan: @melissa what's the right thing to do with mistaken packages?

melissa: Well, other then not ship them in the first place (sorry, most likely my bad), probably just delete them? I remember @haus having strong opinions on this

source/_includes/pup43_platforms_fedora.markdown
gguillotte: OK, but I'm afraid I don't follow; this language isn't new, and if we change it here, we need to reassess how we define "Puppet" and "packages" throughout the docs.

Are you saying we should specify `puppet-agent` and `puppetserver` packages here, or Puppet Collections, or something else? Or that we shouldn't use the word "official"?

gguillotte: In other words, out of these, which is best? Or do none of them work?

> We publish and test official `puppet-agent` packages containing Puppet 4.3 on the following versions of Fedora:

> We publish and test `puppet-agent`, `puppet-server`, and `puppetdb` packages for the following versions of Fedora:

> As of Puppet 4.3, we publish and test packages in the Puppet Collection repository that are compatible with the following versions of Fedora:

source/_includes/pup43_platforms_fedora.markdown
gguillotte: What's up with http://yum.puppetlabs.com/puppetlabs-release-pc1-fedora-20.noarch.rpm, then? Is it no longer being updated? (In which case, does this mean we ship different versions of Puppet to different versions of Fedora and call them all PC1?)

source/_includes/pup43_platforms_fedora.markdown
```diff
@@ -0,0 +1,6 @@
+We publish official `puppet-agent` packages and run automated testing on the following versions for Puppet 4.3:
+
+* Fedora 22
+* Fedora 21
```
gguillotte: Removed Fedora 20 per @melissa

source/_includes/pup43_platforms_fedora.markdown
jtappa: We should probably go back and change the 4.0 ones as well. Technically, we aren't creating the packages for Puppet, we're creating them for puppet-agent. It looks like the OS X includes are the most correct:

"We publish official `puppet-agent` packages for the following Mac OS X versions:" 

Puppet is totally left out of the sentence here, which I think is more correct. The runner up I think would be your other suggestion,

"We publish and test official puppet-agent packages containing Puppet 4.3 on the following versions of Fedora:"

Because of the word "containing", I think this clears up the fact that the package isn't for puppet, it's for AIO.

source/guides/puppetlabs_package_repositories.markdown
```diff
@@ -18,33 +18,37 @@ Puppet Collections are numbered with integers, beginning with PC1. The higher th
...
+    sudo rpm -ivh https://yum.puppetlabs.com/puppetlabs-release-pc1-fedora-22.noarch.rpm
 
 #### Fedora 21
 
-    $ sudo rpm -ivh https://yum.puppetlabs.com/puppetlabs-release-pc1-fedora-21.noarch.rpm
+    sudo rpm -ivh https://yum.puppetlabs.com/puppetlabs-release-pc1-fedora-21.noarch.rpm
+
+#### Fedora 20
+
+    sudo rpm -ivh https://yum.puppetlabs.com/puppetlabs-release-pc1-fedora-20.noarch.rpm
```
gguillotte: This technically still works; should I remove it?

jtappa: I think if it's EOL'd we should probably remove it... I would also think the package should be removed from yum.puppetlabs.com...

source/_includes/pup43_platforms_fedora.markdown
gguillotte: :+1: Thanks @jtappa.

source/_includes/pup40_platforms_redhat_like.markdown
melissa: @underscorgan let's come up with a procedure for this once I'm back in the states

source/guides/puppetlabs_package_repositories.markdown
melissa: Yes, they should be removed. We're putting it off until we can archive those packages somewhere though. So, in the meantime, we aren't going to build any new packages, but the release package and any old packages will just hang out there, abandoned and unloved.

source/_includes/pup43_platforms_fedora.markdown
melissa: Thanks @jtappa! I wasn't trying to incite a whole rewrite at this very moment, but it is something to keep in mind. I've noticed this language in other docs as well, referring to hiera and facter and puppet packages when we don't provide individual packages for any of those project anymore. We make them all available as a part of the puppet-agent package. It's definitely a shift in how we and users are communicating, and everyone's still getting used to it :(

source/_includes/pup43_platforms_fedora.markdown
melissa: Well, PC1 isn't a static thing. We'll add and remove platforms as we get support in and they go end of life. The release package should be removed, since we've officially stopped supporting the platform. The main reason it hasn't is that we *eventually* want to create an archive server to put all the old builds that were publicly available. But, we haven't done that yet :(

source/_includes/pup43_platforms_osx.markdown
```diff
@@ -0,0 +1,5 @@
+We publish official `puppet-agent` packages for the following OS X versions:
+
+* 10.11 El Capitan
```
melissa: is 10.11 out yet? cc/ @heathseals

heathseals: yup.  [shipped](http://downloads.puppetlabs.com/mac/10.11/PC1/x86_64/) with 1.3.0.

source/_includes/pup43_platforms_fedora.markdown
gguillotte: Totally @melissa. If the docs are stale elsewhere, we should address it. I'm raising the flag a little higher here as these includes are re-used in several places across the docs, so these changes will ripple, and I want to make sure those pages are also consistent.

melissa:
Coolio. @gguillotte I haven't looked through this very closely, but you are aware of my main thoughts, so don't let my being distracted prevent this from being merged (I also have very little insight into the docs system, so I'm sure there would be stuff I just don't have context on)

gguillotte:
Tweaked, rebased, and squashed! cc @jtappa

jtappa:
:+1:

PR #580, opened 2015-11-20
Update Puppet Server index page.

gguillotte:
@camlow325 for a quick look before I merge.

### DIFF
--- a/source/puppetserver/index.markdown
+++ b/source/puppetserver/index.markdown
@@ -4,18 +4,17 @@ title: "Puppet Server Documentation"
 toc: false
 ---
 
-
 Puppet Server is the next-generation application for managing Puppet agents.
 
 For the most part, we've integrated the Puppet Server docs into [the Puppet docs.](/puppet/latest/reference) For a list of all docs pages related to Puppet Server, or for info about an older version of Puppet Server, see one of the links below.
 
-
 ## Current Versions
 
-* [Puppet Server 2.1](./2.1) is compatible with Puppet 4.x.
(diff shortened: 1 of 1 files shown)


source/puppetserver/index.markdown
```diff
@@ -4,18 +4,17 @@ title: "Puppet Server Documentation"
...
 
-* [Puppet Server 2.1](./2.1) is compatible with Puppet 4.x.
+* [Puppet Server 2.2](./2.2) is compatible with Puppet 4.3 and newer.
 * [Puppet Server 1.1](./1.1) is included with Puppet Enterprise 3.8 (starting with 3.8.1). It is compatible with Puppet 3.8.
 
 ## Older Versions
 
-- [Puppet Server 2.0](./2.0)
-- [Puppet Server 1.0](./1.0) is included with Puppet Enterprise 3.7 and 3.8.0. It is compatible with Puppet 3.7 and 3.8.
+* [Puppet Server 2.1](./2.1) is compatible with Puppet 4.2 and older.
```
camlow325: This is technically compatible with Puppet 4.3+ as well, although we aren't guaranteed to be testing it this way.  Maybe just say that the 2.0 and 2.1 are compatible with Puppet 4.x?

camlow325: Assuming you want to work compatible PE versions in here as well, could say that Puppet Server 2.1 is compatible with PE 2015.2.x and that Puppet Server 2.2 is "expected to be" compatible with PE 2015.3.x - or maybe just leave off the last part because 2015.3.x has not shipped yet?

PR #584, opened 2015-11-24
Agent 1.3.1 release notes

gguillotte:
Add release notes for Puppet 4.3.1, Facter 3.1.3, and `puppet-agent` 1.3.1, and adds or updates links to `puppet-agent` notes and content to older release notes for Puppet and Facter.

### DIFF
--- a/source/puppet/4.3/reference/release_notes.markdown
+++ b/source/puppet/4.3/reference/release_notes.markdown
@@ -17,6 +17,19 @@ Make sure you also read the [Puppet 4.0 release notes](/puppet/4.0/reference/rel
 
 Also of interest: the [Puppet 4.2 release notes](/puppet/4.2/reference/release_notes.html) and [Puppet 4.1 release notes](/puppet/4.1/reference/release_notes.html).
 
+## Puppet 4.3.1
+
+Released November 30, 2015.
+
+Puppet 4.3.1 is a bug fix release.
+
+* [Fixed in Puppet 4.3.1](https://tickets.puppetlabs.com/issues/?filter=16211)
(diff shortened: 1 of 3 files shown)


gguillotte:
@ahpook @MikaelSmith

source/puppet/4.3/reference/release_notes.markdown
```diff
@@ -17,6 +17,19 @@ Make sure you also read the [Puppet 4.0 release notes](/puppet/4.0/reference/rel
...
+Released November 30, 2015.
+
+Puppet 4.3.1 is a bug fix release.
+
+* [Fixed in Puppet 4.3.1](https://tickets.puppetlabs.com/issues/?filter=16211)
+* [Introduced in Puppet 4.3.1](https://tickets.puppetlabs.com/issues/?filter=16210)
+
+### Bug Fixes: Miscellaneous
+
+* [PUP-5525: Hiera special pseudo-variables breaking with Puppet 4.3](https://tickets.puppetlabs.com/browse/PUP-5525): Puppet 4.3.0 does not initialize Hiera for use with data bindings in the correct order, and does not correctly set special pseudovariables like `calling_module`. This led to lookups not finding values when the special variables were interpolated in hierarchy data paths.
```
nfagerlund: Was this a problem where it worked fine for `hiera()` function lookups but didn't work for automatic class parameter lookups? 

If so, I suggest saying "automatic class parameter lookup" instead of "data bindings". The latter is a confusing/overly-abstract/overly-broad term that we should try to avoid in our docs.

gguillotte: Addressing in PR 587.

PR #586, opened 2015-11-30
Fixed a typo

shichao-an:


nfagerlund:
tooglin'.

Thanks for the catch!

PR #589, opened 2015-12-03
Point Puppet 3 users to the right Puppet Agent doc

gguillotte:
The `about_agent` page notes that `puppet-agent` is for Puppet 4 only,
but doesn't guide Puppet 3 users looking for help. Add a note pointing
those users to relevant Puppet 3 documentation.


nfagerlund:
Makes sense to me.

PR #595, opened 2016-01-02
DOC-2444: Puppet Server config files in nav

nfagerlund:


gguillotte:
:+1: Will merge when back at work.

PR #596, opened 2016-01-06
(DOC-1875) Add note to tagmail docs regarding Puppet Server.

gguillotte:
Add a note to the `tagmail.conf` doc noting that the `puppetlabs/tagmail` modules is required when using Puppet Server, in order to avoid a potential bug. Also note that the built-in processor was removed in PE 3.8 and open source Puppet 4.


nfagerlund:
Well I dunno why I didn't merge this earlier, but let's merge this.

PR #601, opened 2016-01-15
Jan2016 referencepalooza

nfagerlund:
I believe this is now ready to merge, once I've been sanity-checked. 

This is the reference thinning and relocation project tracked in DOC-1455 and DOC-2432.


gguillotte:
@nfagerlund It looks like you're still working on this branch, at least on variable name case. I haven't found any serious problems, but I'll admit a lot of the build scripting is over my head.

The one thing I did catch: URLs in autogenerated content are http instead of https ("or read them online at http://docs.puppetlabs.com/puppet/latest/reference/man/"). This isn't any different from what we do now, but I think we should still use https URLs everywhere we can.

Rest is :+1: by me.

nfagerlund:
@gguillotte Awesome, thank you. I think I'm gonna merge it. 💪🏼 

I think the http -> https thing in the autogenerated docs is actually mostly resolved... for the upcoming Puppet 4.4 and up. (Yup, just checked out master and grepped for that phrase and the URL is updated.) There was a big long PR for that in the Puppet source, but it was aimed at master instead of stable.

PR #602, opened 2016-01-15
Add a line-height to table cell contents.

gguillotte:
Addresses DOC-2483 (tight text and overlapping preformatted text in tables).

### DIFF
--- a/source/files/stylesheets/style.css
+++ b/source/files/stylesheets/style.css
@@ -103,6 +103,11 @@ table td {
   color: rgba(48,44,109,.9);
 }
 
+table td,
+table th {
+  line-height: 1.6;
+}
+
 table th:first-child {
   -webkit-border-top-left-radius: 0.5em;
   -moz-border-radius-topleft: 0.5em;


nfagerlund:
1. Make sure to specify units in `em`. (`1.6em` has the same visual effect as `1.6` on my browser, but let's protect ourselves.) 
2. 1.6 might actually be a touch too wide. What do you think about 1.45? (We used to get complaints about fluffy tables in the past, and going too far might push it back into annoying territory.)
3. We could also help reduce fluffiness by changing the padding in each cell from 0.5em to 0.25em. (Or in-between.) Anyone have thoughts?

gguillotte:
1.  `line-height` is an exception to sizing rules. Length and percentage `line-height` measurements have poor inheritance properties, and the CSS spec defaults `line-height` to a [unitless value of 1.2](https://www.w3.org/TR/CSS2/visudet.html#propdef-line-height). Mozilla's web development documentation [recommends unitless values for `line-height`, and provides examples](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values).
2.  With a `line-height` less than 1.6 (still narrower than the body `line-height` of 1.75), the boxes that wrap code elements begin to overlap when code text wraps or when two code elements are on adjacent lines within a single table cell.

    1.5 `line-height` on `td`, 2px `padding` on `code` elements:
    1.5 line-height, 2px padding

    1.45 `line-height`, 2px `padding` on `code`:
    1.45 line-height, 2px padding

    If we want a tighter `line-height` without colliding code borders, we should reduce or eliminate the box or border around code elements, or reduce the size of text in code elements.

    1.5 `line-height`, 1px `padding` on `code`:

    1.5 line-height, 1px padding

3.  Reducing the cell padding would bring the boxes around code elements closer to the cell edges, including text in horizontally adjacent table cells, regardless of the line height.

    1.5 `line-height`, 2px `padding` on `code`, 0.3em `padding` on `table tr`:
    1.5 line-height, 2px padding, 0.3 cell padding

nfagerlund:
Oh no kidding?! I stand corrected, had no idea line-height was weird. 

And your demos make a good case for not going any narrower than 1.6 or shrinking the cell padding. (I saw different output, but that could have been a result of mistakenly using ems, or it could be a browser issue.) I'm now OK with merging this as-is.

gguillotte:
:+1: I'll merge it, then. If we get feedback that the tables are too airy, we can revisit reducing the top and bottom cell padding and removing or reducing the border and padding on monospaced text.

PR #612, opened 2016-02-12
(maint) Fix the mco version in puppet-agent 1.3.3

kylog:


nfagerlund:
OK, since we decided that was a burnt tag, I deleted the whole row instead. And I modified the version table generator — now it has a config file where we can add versions we want suppressed.

PR #614, opened 2016-02-15
Manage redirects and improve deployments

danielparks:
This adds a redirect configuration file, so that redirects don't have to be managed through pull requests on puppetlabs-modules.

This requires changes to puppetlabs-modules: https://github.com/puppetlabs/puppetlabs-modules/pull/5346

Deployment process
------------------

This changes the deployment process to make it faster, more reliable, and less likely to be noticed by users. An outline of the new steps:

 1. rsync assets to stage
 2. rsync stage to tarball source directory
 3. create the tarball
 4. move the tarball into stage
 5. swap stage and final (site is now live)
 6. check the NGINX rewrite configuration file
 7. reload NGINX if the configuration file is valid

This process reduces the amount of time during which the live site is changing. Swapping the two directories is nearly instantaneous, but rsyncing takes quite a while.

This also means that the tarball will be readily as soon as the new site is made live.

Directory structure
-------------------

This changes the directory structure of the deployment target. There is now a containing directory, e.g. /var/www/docs.puppetlabs.com, with final and stage subdirectories, and an nginx_rewrite.conf file.

The staging directory for the tarball is now also in the containing directory, instead of ~docsdeploy.

No longer syncing modification times
------------------------------------

For background, Jekyll updates the modification time of pretty much every file when it builds. Rsync uses those modifications times to determine what files have changed, so it thinks every file has changed.

This switches to using fast checksums in rsync, which ignores the modification times, and doesn't update them on the server when they change on the machine the deployment is running from.

The primary benefit is that deployments run more than 5 times faster.

A side benefit of this is that timestamp based caching will work better for users. It's hard for me to tell how useful that would be, but roughly 2% of requests over the last 6 hours on the production web site were "304 Not Modified", which indicates caching was in use.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -168,7 +168,6 @@ task :generate do
 
   system("mkdir -p #{OUTPUT_DIR}")
   system("rm -rf #{OUTPUT_DIR}/*")
-  system("mkdir #{OUTPUT_DIR}/references")
   jekyll()
 
   Rake::Task['symlink_latest_versions'].invoke
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -33,15 +33,14 @@ desc "Release the documentation site"
 remote_task :release do
   Rake::Task['check_build_version'].invoke
   puts "DEPLOYING TO: #{domain}"
-  tarball_name = "puppetdocs-latest.tar.gz"
-  staging_dir = "~/puppetdocs_deploy"
 
-  sh "rsync -av --delete output/ #{domain}:#{deploy_to}/"
+  stage_dir = deploy_to + '/stage'
 
-  run "rm -rf #{staging_dir}"
-  run "cp -R #{deploy_to} #{staging_dir}"
-  run "cd #{staging_dir} && ruby ./linkmunger.rb && tar -czf #{tarball_name} *"
-  run "mv #{staging_dir}/#{tarball_name} #{deploy_to}/#{tarball_name}"
+  sh "rsync -crlpv --delete --force output/ #{domain}:#{stage_dir}/"
+
+  # Create tarball, move everything into place, and reload NGINX. Rake has
+  # trouble mixing stdout and stderr, so we combine them on the remote host.
+  run "sh #{stage_dir}/install.sh #{deploy_to} 2>&1"
 end
 
 end
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -13,18 +13,18 @@ task :mirror1 do
 end
 
 task :preview1 do
-    set :domain,    "#{user}@docspreview1.puppetlabs.lan"
-    set :deploy_to, "/opt/docspreview1"
+    set :domain,    "#{user}@staticweb1-dev.puppetlabs.com"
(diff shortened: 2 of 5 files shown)


danielparks:
Even more text in the [OPS-7555 ticket](https://tickets.puppetlabs.com/browse/OPS-7555).

nfagerlund:
Oh man, I'm excited about the improved rsync speed. :+1: for finding the right option in that massive man page. 

Looking at the other stuff now.

nfagerlund:
Linkmunger updates look fine. I've never benchmarked the Pathname class; is that where a lot of the slowness was coming from? Or was it more about recompiling that regex for every page?

nfagerlund:
@danielparks For the new docspreview hosts: Are you basically ready to throw the switch on that? And are you going to want to redirect docspreview#.puppetlabs.lan to the new hostnames?

nfagerlund:
@danielparks Okay, I looked this over, and I think the process all looks good. My only outstanding questions are:

* The server-side scripting is split between this repo and the site/puppetlabs module — we get `install.sh`, and you get `swap_nginx_conf`. Do you think it'd be easier to keep the two files together, since they're so related and so closely tied to the state of our infrastructure? I'd be cool with putting both in either puppet-docs or puppetlabs-modules, if you think it'd make sense.
    * OTOH, I guess this stuff is all pretty entangled anyway? 
* I don't fully understand nginx directive scoping — does using `include` in the vhost make those un-scoped directives in the rewrite.conf file apply to just that vhost? (I'm assuming you ran this and it works fine; mostly just curious for my own education.)
*

danielparks:
> I've never benchmarked the Pathname class; is that where a lot of the slowness was coming from? Or was it more about recompiling that regex for every page?

It was all Pathname; the regular expression is really fast.

> For the new docspreview hosts: Are you basically ready to throw the switch on that? And are you going to want to redirect docspreview#.puppetlabs.lan to the new hostnames?

Yeah, those are good to go — they're up and running now. Though, if you want to change their names or whatever, that's really easy.

It's up to you about making a redirect. I'm happy to do it.

> The server-side scripting is split between this repo and the site/puppetlabs module — we get `install.sh`, and you get `swap_nginx_conf`.

I appreciate a real m-dash!

Hm. Good call; I'll move `swap_nginx_conf` over into this repo.

> I don't fully understand nginx directive scoping — does using `include` in the vhost make those un-scoped directives in the rewrite.conf file apply to just that vhost?

Yup. Essentially the contents of the included file are just copied into the including file, i.e. the included code has the same scope as the include call.

danielparks:
All right! I combined `swap_nginx_conf` into `install.sh`.

Please feel free to rebase, edit, or do whatever you'd like with the code or my commit messages. Or ask me to do it. I was amused by you fixing "noop" since it's become the default spelling for me, kind of like "referer." :)

danielparks:
And… I moved the `error page` directive here, too.

This now all outside of the `location /` block, since that doesn't really add anything. You can add blocks and other NGINX configuration as you wish.

PR #616, opened 2016-02-23
Docs home page revisions

gguillotte:
The heights of the two sections at the center of the docs home page
don't always match. The CSS already implements these divs as CSS tables,
but because the borders are built on divs nested inside the cells
instead of the cells themselves, the borders don't take advantage of
this configuration to automatically align when the box heights differ.

This PR revises the home page HTML and CSS to apply the borders to
the table-cell divs instead of the inner boxes, and adds border-spacing
to recreate the gap between boxes. As a result, the site-width, header,
and sub-footer divs also must be adjusted to accomodate the different
alignment produced by the use of border-spacing instead of padding.

Also, wrap the sub-footer in a similar box as the columns, create
three evenly spaced columns instead of a horizontal list of links, and
expand the sub-footer per MKOPS-1500.

### DIFF
--- a/source/files/stylesheets/homepage.css
+++ b/source/files/stylesheets/homepage.css
@@ -148,7 +148,7 @@ body {
 
 /* homepage-sub-footer */
 .homepage-sub-footer {
-  margin-top: 40px;
+  margin-top: 20px;
   text-align: center;
 }
 .homepage-sub-footer .hsf-link {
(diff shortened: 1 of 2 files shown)


jtappa:
:+1: Tested on Chrome, Firefox, IE, Safari. Validated HTML and CSS with W3C validators (which btw we have a ton of errors and warnings, but GG's revision didn't cause any *new* ones to come up so I say pass).

jtappa:
Sorry, clicked close instead of merge on the first stab, I was too excited.

PR #619, opened 2016-03-01
Improve version picker redirects for agent install docs.

gguillotte:
DOC-2583: Customers are hitting 404s when looking for Puppet Agent install docs from Puppet 3.8 Linux platform docs using the version picker. Redirect them from Puppet 4 picker links to the corresponding pre-install docs, as we already suggest they do at the top of the content.

### DIFF
--- a/source/nginx_rewrite.conf
+++ b/source/nginx_rewrite.conf
@@ -312,3 +312,9 @@ rewrite /pe/(latest|201[6-9][\d\.]+)/accounts_class.html           https://forge
 rewrite /pe/2015.3/accounts_class.html                             https://forge.puppetlabs.com/puppetlabs/accounts  permanent;
 
 rewrite /pe/(latest|2015\.[3-9]|201[6-9]\.\d+)/install_answer_file_reference.html  /pe/$1/install_complete_answer_file_reference.html permanent;
+
+# DOC-2583: Customers are hitting 404s when looking for Puppet Agent install docs from Puppet 3.8 Linux platform docs using the version picker.
+# Redirect them from Puppet 4 picker links to the corresponding pre-install docs, as we already suggest they do at the top of the content.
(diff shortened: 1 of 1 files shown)


source/nginx_rewrite.conf
```diff
@@ -312,3 +312,9 @@ rewrite /pe/(latest|201[6-9][\d\.]+)/accounts_class.html           https://forge
 rewrite /pe/2015.3/accounts_class.html                             https://forge.puppetlabs.com/puppetlabs/accounts  permanent;
 
 rewrite /pe/(latest|2015\.[3-9]|201[6-9]\.\d+)/install_answer_file_reference.html  /pe/$1/install_complete_answer_file_reference.html permanent;
+
+# DOC-2583: Customers are hitting 404s when looking for Puppet Agent install docs from Puppet 3.8 Linux platform docs using the version picker.
+# Redirect them from Puppet 4 picker links to the corresponding pre-install docs, as we already suggest they do at the top of the content.
+rewrite /puppet/(latest|4.[0-4])/reference/install_(debian_ubuntu|el|fedora|osx|windows).html    /puppet/$1/reference/install_pre.html permanent;
```
gguillotte: `install_gem` and `install_tarball` are intentionally omitted from this regex because I think we don't have equivalent docs for Puppet 4.

gguillotte:
@nfagerlund :wave:

source/nginx_rewrite.conf
nfagerlund: I think we should just match on `4.\d+` — we don't want to have to keep going back to update this.

nfagerlund: Other than that, lgtm.

gguillotte:
Updated per @nfagerlund 's feedback.

PR #620, opened 2016-03-02
Simplify and improve the future_lang redirects.

gguillotte:
Instead of having a separate redirect for each future_lang page, and instead of only redirecting `latest` links, use regex match groups to redirect all future_lang_ pages to specific equivalents, for any Puppet 4 version selected from the version picker. (If there isn't an equivalent, it'll 404 anyway.)


nfagerlund:
Good call.

PR #622, opened 2016-03-05
Add static catalog docs.

gguillotte:
This pull request adds an overview and basic implementation documentation for static catalogs. It depends on [`puppet-server` PR 941](https://github.com/puppetlabs/puppet-server/pull/941) and [`puppet` PR 4749](https://github.com/puppetlabs/puppet/pull/4749).

CC @joshcooper @KevinCorcoran @fpringvaldsen @ipeldridge

### DIFF
--- a/source/puppet/4.4/reference/static_catalogs.md
+++ b/source/puppet/4.4/reference/static_catalogs.md
@@ -49,7 +49,7 @@ Consequently, the agent's Puppet runs might produce different results each time 
 
 Additionally, each time a Puppet agent applies a normal cached catalog that contains file resources sourced from `puppet:///` locations, the agent requests file metadata from the master each time the catalog's applied, even though nothing's changed in the cached catalog. This causes the master to perform unnecessary resource-intensive checksum calculations for each such file resource.
 
-Static catalogs avoid these problems by including metadata that refers to a specific version of the resource's file. This prevents the a newer version from being incorrectly applied, and avoids having the agent regenerate the metadata on each Puppet run.
+Static catalogs avoid these problems by including metadata that refers to a specific version of the resource's file. This prevents the a newer version from being incorrectly applied, and avoids having the agent request the metadata on each Puppet run.
 
 These are the traits that make this type of catalog "static": it contains all of the information that an agent needs to determine whether the node's configuration matches the instructions and **static** state of file resources **at the point in time when the catalog was compiled.**
 
@@ -72,7 +72,7 @@ With static catalogs enabled, the Puppet master generates metadata for each file
 
 Inlined metadata is part of a `FileMetadata` object in the catalog that's divided into two new catalog sections: `metadata` for metadata associated with individual files, and `recursive_metadata` for metadata associated with many files. To use the appropriate version of the file content for the catalog, [Puppet Server][] also adds a `code_id` parameter to the catalog. The value of `code_id` is a unique string, such as the hash of a git or r10k commit, that corresponds to the version of all files in an environment at the time when the catalog was compiled.
 
-When applying a file resource from a static catalog, an agent first checks the catalog for that file's metadata. If it finds some, Puppet uses the metadata to call the [`static_file_content`][] API endpoint on the Puppet master and retrieve the `code_content`. If the catalog doesn't contain metadata for the resource, Puppet does what it's always done: generate the metadata itself, then request the resource's file from the master in the file's current state.
+When applying a file resource from a static catalog, an agent first checks the catalog for that file's metadata. If it finds some, Puppet uses the metadata to call the [`static_file_content`][] API endpoint on the Puppet master and retrieve the `code_content`. If the catalog doesn't contain metadata for the resource, Puppet does what it's always done: request the file resource's metadata, then request the resource's file from the master in its current state.
 
 ### Configuring `code_id` and the `static_file_content` endpoint
 
@@ -114,7 +114,7 @@ The script's standard output is then provided as the file's `code_content`. Agai
 
 ### Enabling or disabling static catalogs
 
-If you're using static catalogs, the agents don't need to generate file metadata or recurse into directories, catalog application time significantly decreases. And since static catalogs allow agents to use static catalogs more reliably, they're less likely to need the master to generate catalogs as frequently.
+If you're using static catalogs, the agents don't need to request file metadata or recurse into directories. And since static catalogs allow agents to use static catalogs more reliably, they're less likely to need to request catalogs as frequently.
 
 In other words, even if you aren't using static catalogs, disabling it doesn't substantially improve server or agent performance.
 
--- a/source/puppet/4.4/reference/static_catalogs.md
+++ b/source/puppet/4.4/reference/static_catalogs.md
@@ -6,6 +6,7 @@ canonical: "/puppet/latest/reference/static_catalogs.html"
 
 [catalogs]: ./subsystem_catalog_compilation.html
 [catalog endpoint]: ./http_api/http_catalog.html
+[`file_content`]: ./http_api/http_file_content.html
 [`static_file_content`]: /puppetserver/latest/puppet-api/v3/static_file_content.html
 [resource_declaration]: ./lang_resources.html
 [file resources]: ./types/file.html
@@ -41,7 +42,7 @@ Unlike a standard Puppet catalog, a static catalog includes metadata that identi
 
 ### Why use static catalogs?
 
-When a Puppet master produces a normal catalog, the result doesn't specify the a version of file resources. When the agent applies the catalog, it always retrieves the latest version of that file resource, or uses a previously retrieved version if it matches the latest version's contents.
+When a Puppet master compiles a normal catalog, the catalog doesn't specify a particular version of its file resources. When the agent applies the catalog, it always retrieves the latest version of that file resource, or uses a previously retrieved version if it matches the latest version's contents.
 
 When a Puppet manifest depends on a file whose contents might change more frequently than the Puppet agent receives new catalogs --- for instance, if the agent is caching catalogs or can't reach a Puppet master over the network --- this can lead to a node applying a version of the referenced file that doesn't match the instructions in the catalog.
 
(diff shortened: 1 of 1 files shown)


source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,130 @@
...
+
+### What's a static catalog?
+
+Unlike a standard Puppet catalog, a static catalog includes metadata that identifies the desired state of any [file resources][] on a node that have `source` attributes pointing to `puppet:///` locations. This metadata can refer to a specific version of the file, rather than the latest version, and confirm that the agent is applying the appropriate version of the file resource for the catalog. Also, because much of the metadata is provided in the catalog, Puppet agents make fewer requests to the master.
+
+### Why use static catalogs?
+
+When a Puppet master produces a normal catalog, the result doesn't specify the a version of file resources. When the agent applies the catalog, it always retrieves the latest version of that file resource.
```
KevinCorcoran: The end of this paragraph could give the impression that the agent always fetches a file from the master, even if it's already fetched it in the past and has it on disk.  I might adjust the wording to avoid that misunderstanding.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Configuring `code_id` and the `static_file_content` endpoint
+
+When requesting the file's content from a static catalog's metadata, the Puppet agent passes the file's path, the catalog's `code_id`, and the requested environment to Puppet Server's new [`static_file_content`][] API endpoint, and the endpoint returns the appropriate version of the file's contents as the `code_content`.
+
+> **Puppet Enterprise note:** If [file sync][] is enabled on future versions of Puppet Enterprise that include Puppet Server 2.3, Puppet Server automatically populates the `code_id` and serves the `code_content` out of the box. You don't need to read any further to configure static catalogs --- they're already configured. However, if you've disabled file sync (for instance, to use r10k instead of Code Manager) and want to use static catalogs, you'll need to read on for information about creating the necessary scripts and configuring Puppet Server.
```
KevinCorcoran: `However, if you've disabled file sync ...` might need some re-wording now that we've decided to disable file sync by default.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,130 @@
...
+
+If static catalogs are enabled but none of the relevant Puppet Server settings are configured, the `code_id` parameter defaults to a null value and the `static_file_content` API endpoint always returns the latest content. To populate the `code_id` with a more useful identifier and have the `static_file_content` endpoint return the appropriate version of the file's content, you must specify scripts or commands that provide Puppet with the right information.
```
KevinCorcoran: > the `static_file_content` API endpoint always returns the latest content

This is not correct; in this situation, `/static_file_content` is not used at all - the agent simply makes old-style requests to `/file_content` as before.

source/puppet/4.4/reference/static_catalogs.md
ipeldridge: @gguillotte you could just revise that section to point to PE page. I'll cover the new change there. "You don't need to..." could just become a link that says to enable them, etc.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,130 @@
...
+
+Puppet Server locates these commands via the `code-id-command` and `code-content-command` settings in Puppet Server's [`puppetserver.conf`][] file. Puppet Server runs the `code-id-command` each time it compiles a static catalog, and it runs the `code-content-command` each time an agent requests file contents from the `static_file_content` endpoint.
+
+> **Puppet Enterprise note:** You can set `code-id-command` and `code-content-command` in the PE console by changing the `code_id_command` and `code_content_command` parameters. Note that these only take effect if [file sync][] is disabled by setting the `file_sync_enabled` console parameter to false.
```
KevinCorcoran: > Note that these only take effect if [file sync][] is disabled by setting the `file_sync_enabled` console parameter to false.

Probably re-word to adjust for file sync off by default.  Also I think this should probably include some sort of disclaimer about how most users don't want to do this unless they are not using file sync and want static catalogs.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,130 @@
...
+#!/bin/bash
+set -e
+if [[ -z $1 ]]; then
+  echo Expected an environment >&2
+  exit 1
+fi
+cd /etc/puppetlabs/code/environments/$1 && git rev-parse HEAD
+```
+
+As long as the script's standard output is a valid hash, Puppet Server will store it as the catalog's `code_id`. If the script returns a non-zero exit code, Puppet Server logs an error.
```
KevinCorcoran: > As long as the script's standard output is a valid hash

That's not correct.  I'd replace with "As long as the script's exit code is zero, Puppet Server will use the scripts standard output as ..."

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Pinging @joshcooper as this was noted in his Direct Puppet doc.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: :+1:

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,130 @@
...
+#!/bin/bash
+set -e
+if [[ $# < 3 ]]; then
+  echo Expected environment, code-id, file-path >&2
+  exit 1
+fi
+cd /etc/puppetlabs/code/environments/$1 && git show $2:$3
+```
+
+The script's standard output is then provided as the file's `code_content`. Again, if the script returns a non-zero exit code, Puppet Server logs an error.
```
KevinCorcoran: I believe this error is also returned in the `/static_file_content` response along with a 500 code.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,130 @@
...
+if [[ -z $1 ]]; then
+  echo Expected an environment >&2
+  exit 1
+fi
+cd /etc/puppetlabs/code/environments/$1 && git rev-parse HEAD
+```
+
+As long as the script's standard output is a valid hash, Puppet Server will store it as the catalog's `code_id`. If the script returns a non-zero exit code, Puppet Server logs an error.
+
+A `code-content-command` script can also be simple. The following example assumes the environment name is passed to the script as the first parameter, the `code_id` is passed as the second parameter, and the path to the file resource from its `content_uri` is passed as the third parameter:
```
KevinCorcoran: Instead of "The following example assumes ...", I'd probably phrase this as something more like "Code content scripts will get the environment name passed as the first command-line argument, ..."

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Between a rock and a hard place: Ideally this would link back to the PE docs covering this subject, but they won't exist when 2.3.0's released. The disclaimer would be moot when 2.3.0 comes out anyway, and would be covered by the PE docs when PE includes this feature. There'll be a follow-on PR once that's the case.

Alternatively, we can just pull all the PE-specific stuff from this PR and put it in that follow-on PR.

gguillotte: Sounds like we'll remove the PE-specific stuff for now, and I'll file a follow-on restoring it after tech and developmental reviews are done. CC @ipeldridge @kayayarai

gguillotte:
Incorporated feedback from @KevinCorcoran and removed PE content per @ipeldridge.

source/puppet/4.4/reference/static_catalogs.md
joshcooper: My doc is incorrect, I'll add a card to update that. If there's no `content_uri` in the file metadata, then the agent will fall back to using `file_content`

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,124 @@
...
+
+Puppet Server locates these commands via the `code-id-command` and `code-content-command` settings in Puppet Server's [`puppetserver.conf`][] file. Puppet Server runs the `code-id-command` each time it compiles a static catalog, and it runs the `code-content-command` each time an agent requests file contents from the `static_file_content` endpoint.
+
+Puppet Server validates the standard output of each of these scripts, and if the output's acceptable, it adds the results to the catalog as their respective parameters' values. This lets you use any versioning or synchronization tools you want, as long as you can write scripts that produce a hash for the `code_id` and code content using the catalog's `code_id` and file's environment.
```
KevinCorcoran: I think we should use a word other than "hash" here, as readers might not know if this means a SHA-1 hash for a Git commit ID or a Ruby hash which is totally different.  Maybe just "string" or "unique identifier" or something like that.

KevinCorcoran:
one last nitpick, otherwise :shipit:

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,124 @@
...
+A `code-content-command` script can also be simple. Puppet Server passes the environment name as the first command-line argument, the `code_id` as the second argument, and the path to the file resource from its `content_uri` as the third argument:
+
+``` bash
+#!/bin/bash
+set -e
+if [[ $# < 3 ]]; then
+  echo Expected environment, code-id, file-path >&2
+  exit 1
+fi
+cd /etc/puppetlabs/code/environments/$1 && git show $2:$3
```
joshcooper: It might be good to have someone review these with an eye for security related issues, e.g. should `$1`, etc be quoted?

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+The script's standard output is then provided as the file's `code_content`. Again, if the script returns a non-zero exit code, Puppet Server logs an error and returns the error and a 500 response code to the API request.
+
+### Enabling or disabling static catalogs
+
+If you're using static catalogs, the agents don't need to generate file metadata or recurse into directories, catalog application time significantly decreases. And since static catalogs allow agents to use static catalogs more reliably, they're less likely to need the master to generate catalogs as frequently.
```
joshcooper: Might want to change "generate" to "request". The agent still generates local file metadata (when static catalogs are enabled). But the question is whether the agent compares it against the metadata in the catalog, or if the agent needs to request the metadata from the server.

When static catalogs are enabled, I'm not sure that application time will decrease significantly... The agent and master are doing the same amount of work. The main difference is the agent gets all of its questions answered at once.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,124 @@
...
+
+In other words, even if you aren't using static catalogs, disabling it doesn't substantially improve server or agent performance.
+
+However, if you still want to toggle static catalog generation, you can do so with the Boolean `static_catalog` setting in two places:
```
joshcooper: It's plural [`static_catalogs`](https://github.com/puppetlabs/puppet/blob/stable/lib/puppet/defaults.rb#L151)

gguillotte:
Re-added WIP due to several changes to static catalogs in the last two days.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: It sounds like a schema for `code-id-command` is [in the pipe](https://github.com/puppetlabs/puppet-server/pull/956/files#diff-6dbd5ce712f90995e8774c3cae74f54bR8), so I'll revise accordingly when that PR is settled.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: @joshcooper The point is to demonstrate how inputs and outputs work, not to provide a fully functional, production-ready workflow. Should that be more explicit?

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Addressed in 4e14013.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Addressed in b6972de.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Addressed in 306808b.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Addressed in 5d160fa.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Addressed in b6972de.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Addressed in b6972de.

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Addressed in 63ebb64 L82.

gguillotte:
Ready for review with new content @KevinCorcoran @joshcooper @ipeldridge

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,133 @@
...
+
+> **Note: The Puppet Server process must be able to execute these scripts. Puppet Server also validates their output and checks their exit codes. Environment names can contain only alphanumeric characters and underscores (`_`). The `code_id` can  contain only alphanumeric characters and dashes (`-`), underscores (`_`), semicolons (`;`), and colons (`:`). If either command returns a non-zero exit code, Puppet Server logs an error and returns the error message and a 500 response code to the API request.
+
+Puppet Server validates the standard output of each of these scripts, and if the output's acceptable, it adds the results to the catalog as their respective parameters' values. This lets you use any versioning or synchronization tools you want, as long as you can write scripts that produce a hash for the `code_id` and code content using the catalog's `code_id` and file's environment.
```
KevinCorcoran: This still uses the word `hash`.  Sorry, not sure why I pointed this out last time around, but I'm still thinking it's better to avoid that word here.

KevinCorcoran:
always one more nitpick, but :+1: otherwise.

gguillotte:
@KevinCorcoran :hash: obliterated

KevinCorcoran:
:shipit:

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,132 @@
...
+
+Additionally, each time a Puppet agent applies a normal cached catalog that contains file resources sourced from `puppet:///` locations, the agent requests [file metadata][] from the master each time the catalog's applied, even though nothing's changed in the cached catalog. This causes the master to perform unnecessary resource-intensive checksum calculations for each such file resource.
+
+Static catalogs avoid these problems by including metadata that refers to a specific version of the resource's file. This prevents the a newer version from being incorrectly applied, and avoids having the agent request the metadata on each Puppet run.
+
+These are the traits that make this type of catalog "static": it contains all of the information that an agent needs to determine whether the node's configuration matches the instructions and **static** state of file resources **at the point in time when the catalog was compiled.**
+
+### Static catalog features
+
+Puppet 4.4 and Puppet Server 2.3.0 enable static catalog generation by default, whether you upgrade Puppet or perform a clean installation.
```
joshcooper: This is a little tricky. The `static_catalogs` setting is enabled by default, but the overall behavior is disabled by default, since users need to provide puppetserver `code-*-command` settings and scripts.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,132 @@
...
+A `code-content-command` script can also be simple. Puppet Server passes the environment name as the first command-line argument, the `code_id` as the second argument, and the path to the file resource from its `content_uri` as the third argument:
+
+``` bash
+#!/bin/bash
+set -e
+if [[ $# < 3 ]]; then
+  echo Expected environment, code-id, file-path >&2
+  exit 1
+fi
+cd /etc/puppetlabs/code/environments/"$1" && git show $2:$3
```
joshcooper: Should `$2` and `$3` be quoted too? /cc @haus

haus: `$3` should, because the path may contain spaces or other special chars. It couldn't hurt to quote `$2` as well IMO.

source/puppet/4.4/reference/static_catalogs.md
```diff
@@ -0,0 +1,132 @@
...
+The script's standard output is then provided as the file's `code_content` as long as the script returns a non-zero exit code.
+
+### Enabling or disabling static catalogs
+
+If you're using static catalogs, the agents don't need to request [file metadata][] or recurse into directories. And since static catalogs allow agents to use static catalogs more reliably, they're less likely to need to request catalogs as frequently.
+
+In other words, even if you aren't using static catalogs, disabling it doesn't substantially improve server or agent performance.
+
+However, if you still want to toggle static catalog generation, you can do so with the Boolean `static_catalogs` setting in two places:
+
```
joshcooper: It might be useful to call out situations where puppet will not produce a static calalog: 

1. If the `static_catalogs` puppet setting is false (either globally or in a per-environment `environment.conf`)
1. If using webrick or passenger
1. If an older agent (pre-4.4.0) makes a catalog request
1. If the code_id writes an empty string to stdout (I think this is the case?)

Additionally, puppet will not inline metadata for individual file resources if the resource:

1. Doesn't use the `source` parameter (and uses `content` instead).
1. Uses a source parameter with a non-puppet scheme, e.g. `source => 'http://host:port/path/to/file'`
1. Uses a source parameter that doesn't use the built-in `modules` mount point
1. Uses a source parameter, but the file on the master is not in `/etc/puppetlabs/code/environments/<environment>/*/*/files/**`. For example, module files are typically in `/etc/puppetlabs/code/environments/<environment>/modules/<module_name>/files/**`

gguillotte: @joshcooper I don't understand point 3. Doesn't point 4 contradict that? Or do you mean Puppet Server will inline metadata if either point 3 or point 4 is true?

joshcooper: Sorry I think I mixed positive and negatives in my points. Let me update those.

Longer story, puppet has a built-in `modules` mount point (along with `plugins`, `pluginsync`, etc). Users can also create their own mount points to serve other files. So point 3 is to make sure the request was for the `modules` mount point specifically.

If the request gets past that check, then we try to resolve where the actual file is on the master (to compute checksum, etc). Normally the pattern is a request for `source => 'puppet:///modules/foo/bar'` will resolve to `/etc/puppetlabs/code/environments/<environment>/modules/foo/files/bar`, and in that case we would inline the resource.

However, due to symlinks (assuming opensource puppetserver with no file sync), it's possible that the file lives outside of the per-environment directory. So that's a case where we'd get past point 3, but fail on 4, and not inline metadata.

Also it's possible to have an `environment.conf` that [modifies the `modulepath`](https://github.com/puppetlabs/control-repo/blob/production/environment.conf#L1). If the resolved path is something like `/etc/puppetlabs/code/environments/<environment>/dist/foo/files/bar` then we would inline, even though `modules` is not in the file path.

It's also possible for `modulepath` in `environment.conf` to be something like `/usr/modules`, and in that case we would not inline.

gguillotte:
More feedback assimilated, ready for another review. CC @joshcooper @KevinCorcoran @ipeldridge

source/puppet/4.4/reference/static_catalogs.md
gguillotte: Awesome description, thanks. That makes sense now. I've tweaked the text further to try to reinforce that.

joshcooper:
:+1:

PR #625, opened 2016-03-17
(PDB-2502) Prepare for release 4.0 of PuppetDB

kbarber:
This patch switches the branching over for 3.2 to 3.2.x, and 4.0 to
stable, but locks the latest version to be 3.2 for now, until we release.

Signed-off-by: Ken Barber <ken@bob.sh>

### DIFF
--- a/source/_config.yml
+++ b/source/_config.yml
@@ -56,6 +56,7 @@ symlink_latest:
   - facter
   - pe
 lock_latest:
+  puppetdb: 3.2
 #  puppet: 4.3
 #  hiera: 3.0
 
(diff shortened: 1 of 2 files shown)


kbarber:
@puppetlabs/documentarians @nfagerlund @MFredette hi there, could someone please merge this in, as we're releasing today.

Its safe to merge now as its mainly preparation work, but it will help us get a particular version for our soon to be release 4.0, and we'll get 3.2 pointing at 3.2.x now (the branch is already done). We'll follow up with a removal of the lock_latest after this is merged and we have released.

source/_config.yml
```diff
@@ -387,13 +388,21 @@ documents:
       commit: ce030bb
       subdirectory: documentation
 
+  /puppetdb/4.0
```
nfagerlund: typo! Needs a colon, since it's a key in a hash!

PR #632, opened 2016-04-20
Reduce dt bold weight to 500.

gguillotte:
CC @nfagerlund @jtappa 

cf. https://github.com/puppetlabs/puppet-docs-bootstrap/commit/480de6feb55e386297e7768596c29d25eaff3f35


nfagerlund:
Closed in favor of commit 70cc52a, which includes this change.

PR #635, opened 2016-04-22
Remove Marketo forms style tweaks

nfagerlund:
This changes the spacing of elements on /downloads/index.html, but in a way
I don't consider detrimental or even particularly noticeable. TBH I think it
improves it. By the way, this form brings in styling of its own via the javascript
that inserts it into the page -- if the authors believed it should be styled in
a particular way, that's the way they'd style it.


nfagerlund:
To mix things up a bit, I'd like Jamie Carey to review this. I'll bring her in via email since I can't find her GH account.

nfagerlund:
Got the :+1: from Jamie! Get this gunk outta here.

PR #637, opened 2016-04-22
Add "Jump to table of contents" link on sm and xs viewports

nfagerlund:
On smaller viewports, the sidebar nav gets pushed down below the page content,
and it can take a maddening amount of scrolling to reach it. This should solve
95% of that problem in a nice, non-overkill way.

I think grouping this link with the document/version switchers should help
show which TOC we're talking about (the document-wide one, not the in-page one).

### DIFF
--- a/source/_layouts/default.html
+++ b/source/_layouts/default.html
@@ -55,7 +55,9 @@
 
     <div id="doc-switchers" class="container">
       <div class="col-md-8">{% include template/doc_switcher.html %} {% include template/version_switcher.html %}</div>
-      <div class="col-md-4"><div class="visible-sm-inline visible-xs-inline"><a href="#document-nav">Jump to table of contents</a></div></div>
+      <div class="col-md-4"><div class="visible-sm-inline visible-xs-inline"><a href="#document-nav">
+        Jump to {% if page.doc %}{{ site.document_names[page.doc.doc] }} navigation{% else %}related pages{% endif %}
+      </a></div></div>
     </div>
 
     <div class="container">
--- a/source/_layouts/default.html
+++ b/source/_layouts/default.html
@@ -55,7 +55,7 @@
 
     <div id="doc-switchers" class="container">
       <div class="col-md-8">{% include template/doc_switcher.html %} {% include template/version_switcher.html %}</div>
-      <div class="col-md-4"></div>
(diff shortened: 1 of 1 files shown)


source/_layouts/default.html
```diff
@@ -55,7 +55,7 @@
 
     <div id="doc-switchers" class="container">
       <div class="col-md-8">{% include template/doc_switcher.html %} {% include template/version_switcher.html %}</div>
-      <div class="col-md-4"></div>
+      <div class="col-md-4"><div class="visible-sm-inline visible-xs-inline"><a href="#document-nav">Jump to table of contents</a></div></div>
```
gguillotte: Since this jumps to the document nav instead of the page nav, "table of contents" seems ambiguous. Could this use a new variable to be more specific, like 

`Jump to {{product.title}} menu`

or

`Jump to {% if page.url contains "/puppetserver/" %}Puppet Server{% elif page.url contains "/puppetdb/" %}PuppetDB{% endif %} menu` ... etc.

Also, to mirror the "Back to top" link at the bottom and indicate what this does, could this include a downward arrow symbol (↓ `&#8595;`) instead of "Jump to"? That could make it a little more localization friendly, too.

nfagerlund: Hmmmmmmm. Yeah, that's all super feasible. But we'll also need an option to use when we're not in a named product manual, since there's still a sidebar nav to jump to. 

(Bah, the double-nav problem comes back to bite us in the butt again.)

Do you think it would be simpler to refer to it as the "sidebar navigation?" I realize that's not strictly true on mobile (which is the whole problem here), but I wonder if that would actually be the least confusing approach.

gguillotte: What about "product navigation", "documentation outline", or even just "other pages"? Something more descriptive of what it contains, as opposed to where it might be located in the layout.

nfagerlund:
So that updated bit of template code results in stuff like the following: 

* "Jump to Puppet Enterprise navigation" on a page in the PE manual
* "Jump to related pages" on a page like https://docs.puppet.com/background/ssl/public_key.html

Which I think seems like a good stopping point for now. Yeah?

gguillotte:
@nfagerlund LGTM. :shipit:

PR #638, opened 2016-04-22
Remove or modify references to Server's puppet-admin settings.

gguillotte:
Supporting DOC-2615. CC @nfagerlund

### DIFF
--- a/source/puppet/4.4/reference/config_important_settings.markdown
+++ b/source/puppet/4.4/reference/config_important_settings.markdown
@@ -78,6 +78,7 @@ canonical: "/puppet/latest/reference/config_important_settings.html"
 [configuring_timeout]: ./environments_configuring.html#environmenttimeout
 [puppetserver_config_files]: {{puppetserver}}/configuration.html
 [settings_diffs]: {{puppetserver}}/puppet_conf_setting_diffs.html
+[auth.conf]: {{puppetserver}}/config_file_auth.html
 [puppet_admin]: {{puppetserver}}/config_file_puppetserver.html
 [jruby_puppet]: {{puppetserver}}/tuning_guide.html#puppet-server-and-jruby
 [jvm_heap_config]: {{puppetserver}}/install_from_packages.html#memory-allocation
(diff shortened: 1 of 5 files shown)


source/puppet/4.4/reference/environments_configuring.markdown
```diff
@@ -1,6 +1,7 @@
 ---
 layout: default
 title: "Configuring Environments"
+canonical: "/puppet/latest/reference/environments_configuring.html"
```
nfagerlund: This is [now auto-discovered](https://confluence.puppetlabs.com/display/DOCS/YAML+Frontmatter#YAMLFrontmatter-canonical:), so we don't need to explicitly write it.

source/puppet/4.4/reference/config_important_settings.markdown
```diff
@@ -171,6 +172,8 @@ Puppet Server has [its own configuration files][puppetserver_config_files]; cons
 * [`jruby-puppet`][jruby_puppet] --- Provides details on tuning JRuby for better performance.
 * [`JAVA_ARGS`][jvm_heap_config] --- Instructions on tuning the Puppet Server memory allocation.
 
+Also, Puppet Server uses its own [auth.conf][] file by default instead of its `puppet-admin` section to authorize client requests.
```
nfagerlund: I think you should just replace that first bullet point instead of writing an addendum — change it to link to the Puppet Server auth.conf page, and say that you have to add auth rules in order to use the admin interface (and maybe say the admin interface is most useful for refreshing environments). 

Less confusing?

nfagerlund:
I noticed something else while I was in there: the auth.conf (legacy) page needs a major update of its above-the-fold text. I'll deal with that in a separate commit, though.

source/puppet/4.4/reference/config_important_settings.markdown
gguillotte: That was my first idea, but I changed my mind since the deprecated method is still allowed. If we're cool with removing the deprecated stuff, that works for me. :+1:

nfagerlund: Yeah cool, let's do it! B/c the purpose of this page is just to alert people to the most important settings in the maelstrom of useless stuff, right? So we might as well delete the deprecated ones and just point to the best approach.

gguillotte:
Updated, rebased to master, and made this change back to Puppet 4.2.

PR #640, opened 2016-04-23
Note removed Util modules.

gguillotte:
Supports DOC-2401.

Looks like BBEdit also stripped some trailing spaces (3.0 L79, 93; 3.1 L34).

CC @jtappa @kylog @MikaelSmith

### DIFF
--- a/source/facter/3.0/release_notes.md
+++ b/source/facter/3.0/release_notes.md
@@ -76,7 +76,7 @@ When running the MCollective agent plugin on Windows, Facter could report a miss
 
 * [FACT-1125](https://tickets.puppetlabs.com/browse/FACT-1125)
 
-### REGRESSION FIX: Resolved an OLE COM Initialization Issue When Reporting Windows Facts 
+### REGRESSION FIX: Resolved an OLE COM Initialization Issue When Reporting Windows Facts
 
 Facter 3.0.0 can fail when reporting Windows facts that require `win32ole`, resulting in a "fail: OLE initialize" error. Facter 3.0.2 resolves this issue.
 
@@ -90,7 +90,7 @@ If a command returns a non-zero exit code or can't be found, Facter 3.0.0 return
 
 ### REGRESSION FIX: Fixed `stderr` Redirection
 
-When Facter 3.0.0 invokes commands to report facts that result in `stderr` output, Facter reports the `stderr` output as the fact. Also, if the required command doesn't exist, Facter 3.0.0 reports a null value. 
+When Facter 3.0.0 invokes commands to report facts that result in `stderr` output, Facter reports the `stderr` output as the fact. Also, if the required command doesn't exist, Facter 3.0.0 reports a null value.
 
 Facter 3.0.2 correctly reports facts in these situations. To view additional diagnostic information, including the invoked commands' `stderr` output, run Facter with the `--debug` flag.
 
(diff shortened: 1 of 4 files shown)


source/facter/3.0/release_notes.md
```diff
@@ -235,6 +235,18 @@ We changed the value of the hardware fact (`os.hardware`) from "x64" to "x86_64"
 
 * [FACT-610](https://tickets.puppetlabs.com/browse/FACT-610)
 
+### BREAK: Removed `file_read.rb` and `registry.rb`
+
+Facter 3 removed the [`file_read.rb`](https://github.com/puppetlabs/facter/blob/2.x/lib/facter/util/file_read.rb) and [`registry.rb`](https://github.com/puppetlabs/facter/blob/2.x/lib/facter/util/registry.rb), which provide the `Facter::Util::FileRead` and `Facter::Util::Registry` modules to Facter. This can break custom facts that rely on these modules, and produce error messages like this:
```
kylog: IIUC the issue isn't specific to these two files. E.g. look at https://github.com/puppetlabs/facter/tree/2.x/lib/facter/util, none of which exist in facter 3. Thus, the reimplementation of facter from ruby to C++ means custom facts require'ing any individual file in the previous implementation will fail. 

In addition, whereas facter 2 effectively exposed almost its entire implementation as if it were public API for authors of custom facts, facter 3 supports a well-considered API: https://github.com/puppetlabs/facter/blob/master/Extensibility.md#custom-facts-compatibility. Thus, custom facts using facter 2 "API"s outside the facter 3 supported set will fail.

With that as background, I'd suggest that:
* we document the APIs in that Extensibility.md link
* this section be reworked to instruct custom facts authors not to require individual implementation files and to prefer only the documented APIs

Since that first point may take a bit of time we could just link to the Extensibility.md for the time being.

And then that second point could then be enhanced to say something like: if you are porting custom facts from facter 2 that rely on old APIs, a workaround would be to copy the facter 2 implementation into the module's `lib/facter` directory as  described below.

Hope that makes sense. Happy to chat more about this in meatspace.

Also /cc @peterhuene for his thoughts.

gguillotte:
Thanks, @kylog! I've revised the notes. CC @jtappa @peterhuene

source/facter/3.0/release_notes.md
```diff
@@ -235,6 +235,18 @@ We changed the value of the hardware fact (`os.hardware`) from "x64" to "x86_64"
...
+
+```
+Error: Facter: error while resolving custom facts in /opt/configmgmt/environments/production/modules/custom_facts/lib/facter/customfact.rb: cannot load such file -- facter/util/file_read
+```
+
+As a workaround, you can rewrite your custom facts to avoid requiring these Ruby implementations, and instead use the [new API](https://github.com/puppetlabs/facter/blob/master/Extensibility.md#custom-facts-compatibility). You can also [copy the necessary Facter 2 Ruby libraries](https://github.com/puppetlabs/facter/tree/2.x/lib/facter/util) to the `lib/facter/util/` subdirectory of the affected `custom_facts` path, such as `/opt/configmgmt/environments/production/modules/custom_facts/lib/facter/util/` in the above example. For more information about writing custom facts for Facter 3, see [the Facter documentation](https://docs.puppet.com/facter/3.0/custom_facts.html).
+
+* [PE-13225](https://tickets.puppetlabs.com/browse/PE-13225)
```
kylog: This was an internal project ticket, so I just moved it to https://tickets.puppetlabs.com/browse/FACT-1400.

petems:
@gguillotte Awesome, I've had this bite a few times with customers using `Facter::Util::IP` in custom facts during an upgrade 👍

gguillotte:
Updated the ticket numbers per @kylog.

CC @jtappa @nfagerlund for a docs review.

source/facter/3.1/release_notes.md
```diff
@@ -282,3 +282,14 @@ The `puppet-agent` package that installs Facter 3.0.0 did not create the `rubysi
 
 * [FACT-1154](https://tickets.puppetlabs.com/browse/FACT-1154)
 
+### BREAK: Removed Ruby Facter implementations
```
nfagerlund: I think this shouldn't be in the Facter 3.1 release notes, since the behavior was changed in 3.0. I'd be :+1: on putting a note in the Facter 3.1 custom facts guides, but :-1: to putting it in the release notes.

nfagerlund:
Other than where to say it, this lgtm.

gguillotte:
- Removed the note from the 3.1 release notes
- At the top of the 3.1 notes, emphasized checking the 3.0 notes for breaks
- Added short versions of the RN to the 3.0 and 3.1 `custom_facts` doc

nfagerlund:
:shipit:?

gguillotte:
@nfagerlund Fine by me

PR #646, opened 2016-05-03
Fix multi-line formatting

danielhoherd:
While reading puppet docs I came across what I assume was a formatting mistake where severable examples were given on one line where a bullet list made more sense. This fixes those examples.

### DIFF
--- a/source/puppet/4.4/reference/configuration.md
+++ b/source/puppet/4.4/reference/configuration.md
@@ -355,7 +355,9 @@ the user that Puppet is supposed to run as, it defaults to a system
 directory, but if it's running as any other user, it defaults to being
 in the user's home directory.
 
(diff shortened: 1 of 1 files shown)


nfagerlund:
That file's auto-generated, so it'll get replaced as soon as we rebuild the references. 

The relevant code is in the Puppet repo, at lib/puppet/reference/configuration.rb around lines 25-35. I suppose since we're replacing the values for those weird dirs anyway, we could just replace them with a multi-line string. 

And now that I'm looking at this, the markup for the default value shouldn't be a list item. That's silly. A bit of bold text would probably look better.

nfagerlund:
Anyway, someone will get at this at some point soon-ish. I reckon making those special values bulleted lists isn't a bad idea.

nfagerlund:
I got around to filing https://tickets.puppetlabs.com/browse/DOCUMENT-545, so I'm closing this PR (since a different code change is needed to adjust this formatting).

PR #659, opened 2016-05-27
(DOCUMENT-558) Forge only supports semver 1.0.0

hunner:
And 2.0.0 is backwards-incompatible so it will not be changed any time in the near future.


gguillotte:
:+1:

PR #656, opened 2016-05-19
Add Wily, Xenial, and F23 PC1 docs.

gguillotte:

PR #661, opened 2016-06-06
Spelling error in Ubuntu Vivid file name

srynot4sale:


gguillotte:
@srynot4sale Thanks for the catch! This typo is already fixed in #656, which is pending re-review, so I'll close this PR. CC @nfagerlund

PR #662, opened 2016-06-07
Remove invalid Hash type example

sathieu:
As stated above:  "you must specify _both_ key type and value type if you're going to specify one of them"


sathieu:
According to PUP-3680, this should be fixed in all docs, since 3.7.4.

PR #663, opened 2016-06-20
(DOCS-2790) Return 410 status with wiki redirect catch-all

danielparks:
This ensures the wiki_redirect.html page never shows up in Google, and anything that redirects to it will fall out of Google too.


nfagerlund:
@pdxmph pre-approved this status code in DOC-2790, so I'm gonna go ahead and merge.

PR #666, opened 2016-07-20
Fix extra space after a backtick.

ody:
This commit removes an extra space after a backtick in the hiera 3.2
  documentation that is causing a link formatting issue on
  docs.puppet.com. First bullet of the "Puppet 4" section of "Using Hiera
  with Puppet".[1]

  https://docs.puppet.com/hiera/3.2/puppet.html#puppet-4


gguillotte:
👍

PR #670, opened 2016-08-05
(PDB-2927) Update PuppetDB doc references for 4.2

senior:


source/_config.yml
```diff
@@ -58,6 +58,7 @@ symlink_latest:
   - facter
   - pe
 lock_latest:
+  - puppetdb 4.1
```
nfagerlund: Syntax error. The value of lock_latest is a YAML hash, not an array of strings.

nfagerlund:
When does your team need this change merged?

source/_config.yml
senior: So what do I need to change it to?

senior:
@nfagerlund it can go anytime, we'll probably get the release out on Monday, which I'll follow this PR up with another one that unpins the latest version.

source/_config.yml
senior: Looks like maybe I swap the [dash for a colon](https://github.com/puppetlabs/puppet-docs/pull/625/files#diff-e6422af2c5a3ea680839278540cfdf29L58)? I'll make the change.

nfagerlund:
Yup, yaml hashes are `key: value`, so that fixed it. I'll merge.

PR #671, opened 2016-08-05
(maint) Include the security@puppet.com email address

kylog:


nfagerlund:
How in the world was this not here before. :|

PR #678, opened 2016-08-11
Add Server 2.5.0

gguillotte:
@nfagerlund I'd appreciate a sanity check/merge on this as I'm not at my desk


nfagerlund:
Looks good to me, and puppet 4.6 already has a dependency on this so I'll just merge it now.

PR #679, opened 2016-08-15
Modernise usage of facts in language references

ahpook:
Prior to this commit, the example code around conditionals and the section on facts did not make use of structured facts.

This commit updates a couple of sections to show examples with modern structured-fact equivalents of old single value facts. It also describes how to access deeply nested fact values using hash syntax.

(Potentially controversial: I've opted to not quote hash keys, because doing so on 'puppet apply' leads to backslash-itis, and I find it visually noisy. AFAIK it is not semantically necessary but this might be making a unilateral style guide call that's not 100% OK)

### DIFF
--- a/source/puppet/4.6/reference/lang_conditional.markdown
+++ b/source/puppet/4.6/reference/lang_conditional.markdown
@@ -30,10 +30,10 @@ Puppet supports "if" and "unless" statements, case statements, and selectors.
 An "if" statement:
 
 ``` puppet
-if $is_virtual {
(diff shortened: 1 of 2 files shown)


nfagerlund:
@ahpook We talked about this in the platform docs sync, and we're gonna go ahead and quote the keys to present a united front with the style guide. Personally I like unquoted keys as well, but we don't want to invite a bunch of "why aren't you following the style guide" PRs. :P

PR #680, opened 2016-08-17
Facts style conformance and backporting

gguillotte:
Followup PR to #679 to quote keys and backport edits to the rest of Puppet 4.

CC @nfagerlund @jtappa @ahpook

### DIFF
--- a/source/puppet/4.1/reference/lang_facts_and_builtin_vars.markdown
+++ b/source/puppet/4.1/reference/lang_facts_and_builtin_vars.markdown
@@ -7,14 +7,14 @@ canonical: "/puppet/latest/reference/lang_facts_and_builtin_vars.html"
 [definedtype]: ./lang_defined_types.html
 [environment]: ./environments.html
 [topscope]: ./lang_scope.html#top-scope
-[core_facts]: /facter/latest/core_facts.html
-[facter]: /facter/latest
-[customfacts]: /facter/latest/custom_facts.html
-[external facts]: /facter/latest/custom_facts.html#external-facts
+[core_facts]: {{facter}}/core_facts.html
+[facter]: {{facter}}
+[customfacts]: {{facter}}/custom_facts.html
+[external facts]: {{facter}}/custom_facts.html#external-facts
 [catalog]: ./lang_summary.html#compilation-and-catalogs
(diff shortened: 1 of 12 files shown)


source/puppet/4.1/reference/lang_facts_and_builtin_vars.markdown
```diff
@@ -79,11 +78,11 @@ All facts appear in Puppet as [top-scope variables][topscope]. They can be acces
 
 Example, with the osfamily fact:
 
-~~~ ruby
-    if $osfamily == 'redhat' {
-      # ...
-    }
-~~~
+``` puppet
+if $osfamily == 'redhat' {
```
ahpook: this one should also be `$facts['os']['family']`

source/puppet/4.1/reference/lang_facts_and_builtin_vars.markdown
```diff
@@ -79,11 +78,11 @@ All facts appear in Puppet as [top-scope variables][topscope]. They can be acces
 
 Example, with the osfamily fact:
```
ahpook: maybe: "Example, using the `family` fact, which aggregates all Red Hat-derived Linux variants into one 'redhat' value."

ahpook:
found one `$osfamily` that wasn't updated, other than that it looks great. Thanks @gguillotte

source/puppet/4.1/reference/lang_facts_and_builtin_vars.markdown
gguillotte: The context here is to explicitly refer to this as a "classic `$fact_name` fact" distinct from the following section, which covers "the `$facts['fact_name']` hash". I might be missing something here, but I'm not sure that replacing this specific $fact_name example with a $facts['fact_name'] example makes sense.

ahpook: ah, doh, that makes sense.  missed it from the line diffs. Ignore me :)

gguillotte:
@jtappa I think that's a 👍  from @ahpook. Not sure if you want to take a look before this gets merged.

jtappa:
:+1:

PR #681, opened 2016-08-24
fix incorrect quoting

sammcj:
add quoting to string used to define group


jtappa:
Nice catch, thanks! 👍

PR #686, opened 2016-09-16
(maint) Update docs to use new signing key

melissa:
We're switching out the gpg key that we're using to sign packages. This
commit updates the docs to reference this new key rather than the old
one

### DIFF
--- a/source/guides/puppetlabs_package_verification.markdown
+++ b/source/guides/puppetlabs_package_verification.markdown
@@ -35,31 +35,32 @@ Before you can verify signatures, you must import the Puppet Labs public key and
 
 To import the release signing key, run:
 
-    $ gpg --recv-key 1054b7a24bd6ec30
-    gpg: requesting key 4BD6EC30 from hkp server pool.sks-keyservers.net
-    gpg: key 4BD6EC30: public key "Puppet Labs Release Key " imported
+    $ gpg --recv-key 7F438280EF8D349F
+    gpg: requesting key EF8D349F from hkp server pool.sks-keyservers.net
+    gpg: key EF8D349F: public key "Puppet, Inc. Release Key (Puppet, Inc. Release Key) <release@puppet.com>" imported
     gpg: no ultimately trusted keys found
     gpg: Total number processed: 1
     gpg:               imported: 1  (RSA: 1)
 
-The key is also [available here](http://pool.sks-keyservers.net:11371/pks/lookup?op=get&search=0x1054B7A24BD6EC30).
+The key is also [available here](http://pool.sks-keyservers.net:11371/pks/lookup?op=get&search=0x7F438280EF8D349F).
 
 ### Verify the Fingerprint
 
 The fingerprint of the Puppet Labs release signing key is **`47B3 20EB 4C7C 375A A9DA  E1A0 1054 B7A2 4BD6 EC30`**. Run the following and ensure the fingerprints match:
 
-    $ gpg --list-key --fingerprint 1054b7a24bd6ec30
-      pub   4096R/4BD6EC30 2010-07-10 [expires: 2016-07-08]
-            Key fingerprint = 47B3 20EB 4C7C 375A A9DA  E1A0 1054 B7A2 4BD6 EC30
-      uid                  Puppet Labs Release Key (Puppet Labs Release Key)
+    $ gpg --list-key --fingerprint 7F438280EF8D349F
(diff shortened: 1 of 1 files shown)


source/guides/puppetlabs_package_verification.markdown
```diff
@@ -35,31 +35,32 @@ Before you can verify signatures, you must import the Puppet Labs public key and
 
 To import the release signing key, run:
```
mckern: I've been trying to update documentation to refer to the private key as the signing key, and the public key as "public key" lately. </nitpick>

source/guides/puppetlabs_package_verification.markdown
```diff
@@ -35,31 +35,32 @@ Before you can verify signatures, you must import the Puppet Labs public key and
...
 
-    $ gpg --recv-key 1054b7a24bd6ec30
-    gpg: requesting key 4BD6EC30 from hkp server pool.sks-keyservers.net
-    gpg: key 4BD6EC30: public key "Puppet Labs Release Key " imported
+    $ gpg --recv-key 7F438280EF8D349F
```
mckern: Do we want to specify the `pgp.mit.edu` keyserver here, with `--keyserver pgp.mit.edu`?

shrug:
👍

source/guides/puppetlabs_package_verification.markdown
```diff
@@ -35,31 +35,32 @@ Before you can verify signatures, you must import the Puppet Labs public key and
...
+    gpg: requesting key EF8D349F from hkp server pool.sks-keyservers.net
+    gpg: key EF8D349F: public key "Puppet, Inc. Release Key (Puppet, Inc. Release Key) <release@puppet.com>" imported
```
mckern: Should we leave a note about this tool possibly printing the short key ID instead of the long one regardless of how it's fetched?

melissa:
cc/ @nfagerlund I'm not sure who else to ping. This change should probably be in both this repo and puppet-docs-private. Should I submit a separate PR against the second repo? Or will this be merged into that one?

nfagerlund:
@jtappa @gguillotte 

* Don't sweat puppet-docs-private. It'll take care of itself.
* I think this is the dead page version of this info. I'm seeing 0 links to this page, but the same info repeated on the following:
    * puppet/3.8/reference/puppet_repositories.md
    * puppet/4.1/reference/puppet_collections.md
    * puppet/4.2/reference/puppet_collections.md
    * puppet/4.3/reference/puppet_collections.md
    * puppet/4.4/reference/puppet_collections.md
    * puppet/4.5/reference/puppet_collections.md
    * puppet/4.6/reference/puppet_collections.md

    And it looks like it's copy-pasted. 
* We should probably do one of the following:
    * Coalesce this stuff into an include file, and then update that single source. 
    * Only update the upcoming Puppet 4.7 docs, since... I THINK we won't be publishing any packages for older versions using the new signing key? Unless we will, in which case things like Puppet 3.8 will need BOTH versions of the verification steps... :/ :/ :? 

So, uh... what next?

gguillotte:
@jtappa @nfagerlund I can shove it into an include.

melissa:
How will puppet-docs-private take care of itself? There are a bunch of additional references to the old key there

nfagerlund:
@melissa 

* The master branch of puppet-docs-private is dead. We never do anything with it. 
* The pe-preview branch of it is only deployed internally, to docspreview1.ops.puppetlabs.net, so its content is never publicly visible. 
* We periodically merge puppet-docs/master into puppet-docs-private/pe-preview, so it'll get the update eventually anyhow.

melissa:
Oh, thanks! Good to know!

nfagerlund:
@melissa No prob! By the way... what's the word on whether we'll ever sign releases in old lines with the new key? Should 3.8 have both versions?

melissa:
Everything in the foss repos will be signed with the new key. PE 3.8 is the only thing that's stuck on the old key, so all PE 3.x docs should keep the old key references

mckern:
@nfagerlund PE 3.8 will die just about the time this key expires; in theory this will be OK.

nfagerlund:
Cool. And... we should only be merging these changes (via Garrett's replacement PR) once the agent 1.7 package is out the door, right?

melissa:
yes! That's when the new key will go live

gguillotte:
@nfagerlund @melissa @mckern Are we OK with closing this in favor of #687?

melissa:
Yes!

PR #689, opened 2016-09-20
(DOC-2937) Fix reference to static_catalogs setting.

gguillotte:
@cprice @joshcooper


joshcooper:
👍

gguillotte:
@joshcooper @cprice404 We're AWOL today as a team so I'll merge now and we can follow on later if necessary

cprice404:
@joshcooper nah that was the bit that threw me.

PR #693, opened 2016-09-23
Clarify Defined Resource Type Example

pizzaops:
The current example for defined resource types is unclear, due to the
coding style used.

In general, listing all defined type parameters on one line is
considered poor style and very difficult to read.

In addition, comments in-line in a resource, between parameters, are
also viewed as poor style, and make the resource harder to read,
especially for new users who are just learning.

Hopefully this refactor clarifies the example.

Even as an "expert" Puppet user, I found the original fairly hard to
read.

### DIFF
--- a/source/puppet/4.7/reference/lang_defined_types.markdown
+++ b/source/puppet/4.7/reference/lang_defined_types.markdown
@@ -36,27 +36,28 @@ Defines can be used as simple macros or as a lightweight way to develop fairly s
 
 You can use a `define` statement to create a new defined resource type.
 
-``` puppet
+```puppet
 # /etc/puppetlabs/puppet/modules/apache/manifests/vhost.pp
 define apache::vhost (
-  integer $port, 
-  string[1] $docroot, 
-  string[1] $servername = $title, 
-  string $vhost_name = '*'
+  Integer $port,
+  String[1] $docroot,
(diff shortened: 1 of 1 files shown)


nfagerlund:
@pizzaops 👍 to the reformatting, thank you. I'll merge once the syntax errors are fixed.

pizzaops:
@nfagerlund it should be fixed now. Sorry about that! Also fixed some minor issues raised by `puppet-lint`.

source/puppet/4.7/reference/lang_defined_types.markdown
```diff
@@ -36,18 +36,26 @@ Defines can be used as simple macros or as a lightweight way to develop fairly s
...
-  include apache # contains Package['httpd'] and Service['httpd']
-  include apache::params # contains common config settings
+define apache::vhost (
+  Integer $port,
+  String[1] $docroot,
+  String[1] $servername = $title,
+  String $vhost_name = '*',
+) {
+  include ::apache # contains package['httpd'] and service['httpd']
+  include ::apache::params # contains common config settings
```
nfagerlund: The leading :: isn't needed for this version of Puppet... but I don't want to keep making you go back and forth. I'll trim that after I merge.

PR #696, opened 2016-10-06
Reference facts via top level hash

rnelson0:
Current best practice is to reference facts via the `$facts` hash rather than as legacy top level variables.


nfagerlund:
👍

PR #697, opened 2016-10-13
(DOCUMENT-495) Fix iteration example code

roman-mueller:
On this page:
https://docs.puppet.com/puppet/latest/reference/lang_iteration.html
The example code is not passing a puppet-lint run.
This PR fixes that.


nfagerlund:
Cool cool.

PR #700, opened 2016-10-27
(DOC-2982) Backport ERB documentation

petems:
- ERB temptlating has been around since older Puppet


jtappa:
The tocs need to be updated for each version you backport, otherwise the only way to get to these pages would be to go to the latest version and then use the version switcher or to stumble upon it in google search.

petems:
@jtappa Cool, will do! 👍

jtappa:
Hey @petems this PR is getting stale, could you update the tocs of the versions you backported, and then I can merge?

petems:
@jtappa Sure, can do. Where are the TOC for 3.X Puppet? I can see 4 uses `_puppet_toc.html` but I can't find that in older versions?

jtappa:
@petems Prior to 4.x, they should be in the _includes directory

petems:
Ok, I've added them to the _includes files, since the layouts changed not sure the best place to put them, can fix if needed! 👍

PR #701, opened 2016-10-28
Add type aliases documentation

domcleal:
Available since Puppet 4.4, so could be copied down to that (and linked from its release notes?) if you wish.

See https://github.com/puppetlabs/puppet-specifications/blob/master/language/expressions.md#type-alias-expression for the specs.


jtappa:
We have an older branch for type aliases that was never merged in. I'm going to close this PR and then merge a version of both of them for docs 4.4 and up.

PR #702, opened 2016-11-01
(maint) fix typo in indirector docs (van->and)

johnmccabe:


jtappa:
🙊  wow, that's been there a while. Thanks for the fix! 🎉

nfagerlund:
@jtappa Heyyyyyyyyyyyyyyyy, this file is auto-generated, so that typo will return from the dead the minute we generate new docs. (It's already in the 4.8 stuff, right?) 

Would you like to handle the PR to core? The file is `lib/puppet/reference/indirection.rb`.

johnmccabe:
@nfagerlund / @jtappa I'll raise this against core now, thx for the heads up

jtappa:
@nfagerlund oh dang, I wasn't even looking at what directory this page was. 😿  Sorry!

PR #708, opened 2016-11-22
(layout) Add download docs link to header

nfagerlund:
screen shot 2016-11-21 at 5 36 33 pm

This adjusts the gap between the two groups of nav items, because there was
a 'dead spot' as you shrank the window where the line would wrap onto two
lines.

Still need appropriate translations for fr and ja.


nfagerlund:
@LarissaLane Hey, could I trouble you for a French and Japanese translation of "Download docs"?

jtappa:
👍

PR #709, opened 2016-12-07
REVIEW before merge: major revision of environment isolation page

jbondpdx:
@KevinCorcoran , @hlindberg and @scotje , could you please give me a tech review on this page?

### DIFF
--- a/source/puppet/4.8/env_isolation.md
+++ b/source/puppet/4.8/env_isolation.md
@@ -8,25 +8,25 @@ description: "Generating metadata to isolate resources in environments in Puppet
 [code_mgr_env]: {{pe}}/code_mgr.html#environment-isolation-metadata
 
 
-If you have multiple environments, environment isolation prevents resource types from leaking between your various environments. To use environment isolation, you'll generate metadata files that Puppet can use in place of the normal Ruby resource type implementations.
+If you have multiple environments, environment isolation prevents module resource types from leaking between your various environments. To use environment isolation, you'll generate metadata files that Puppet can use in place of the normal Ruby resource type implementations.
 
 In open source Puppet, enable environment isolation with the `generate types` command. In Puppet Enterprise, environment isolation is automatically provided by [Code Manager][code_mgr_env]. Environment isolation is not supported for r10k with Puppet Enterprise.
 
 ## Preventing resource types leaks in multiple environments
 
-If you use multiple environments with Puppet, you might encounter issues with multiple versions of the same resource type leaking between your various environments. 
+If you use multiple environments with Puppet, you might encounter issues with multiple versions of the same resource type leaking between your various environments. This happens with custom resource types in modules, not with Puppet's built-in resource types.
 
 This problem occurs because Ruby resource type bindings are global in the Ruby runtime. Thus, the first loaded version of a Ruby resource type takes priority, and then subsequent requests to compile in other environments get that first loaded version.
 
 Environment isolation solves this issue by generating and using metadata that describes the resource type implementation, instead of using the Ruby resource type implementation when compiling catalogs. 
 
-> **Note**: Other environment isolation cases, such as external helper logic issues or varying versions of required gems, are not solved with the `generate type` command.
+> **Note**: Other environment isolation cases, such as external helper logic issues or varying versions of required gems, are not solved with the `generate types` command.
 
 ### Agent logic is unchanged
 
 Catalog compilation results are exactly the same whether you use the metadata-based version of the resource type or the Ruby implementation. The agent acting on the catalog continues to use the Ruby-based implementation. Because the agent uses one, and only one, version of the logic, isolation is not a problem for the agent.
 
-## Enable environment isolation in Puppet
+## Enable environment isolation in open source Puppet
 
 With open source Puppet, enable environment isolation by running the `generate types` command:
 
@@ -38,9 +38,9 @@ For example, to generate metadata for your production environment, run:
 
 Whenever you deploy a new version of Puppet, overwrite previously generated metadata with `puppet generate types --environment <envname> --force`.
 
-### Implement environment isolation with r10k
+### Enable environment isolation with r10k
 
-To use environment isolation with open source Puppet and r10k, generate types for each environment every time r10k deploys new code.
+To use environment isolation with open source Puppet and r10k, generate types for each environment every time r10k deploys new code. (Environment isolation is not supported with r10k in Puppet Enterprise.)
 
 To generate types with r10k, either:
 
@@ -85,6 +85,7 @@ To diagnose or repair issues:
      1. In [Hiera](./config_intro.html#configure-settings-with-hiera), set `puppet_enterprise::master::puppetserver::pre_commit_hook_commands: []`. 
      1. On the command line, run `service pe-puppetserver reload`.
      1. Delete the `.resource_types` directories from your staging code directory (`/etc/puppetlabs/code-staging`).
+     1. Deploy environments.
 
 ## Reference: `generate types`
 
--- a/source/puppet/4.8/env_isolation.md
+++ b/source/puppet/4.8/env_isolation.md
@@ -30,13 +30,24 @@ Catalog compilation results are exactly the same whether you use the metadata-ba
 
 With open source Puppet, enable environment isolation by running the `generate types` command:
 
-1. On the command line, run `generate types --environment <envname>` for each of your environments.
+1. On the command line, run `puppet generate types --environment <envname>` for each of your environments.
 
 For example, to generate metadata for your production environment, run:
 
-`generate types --environment production`
+`puppet generate types --environment production`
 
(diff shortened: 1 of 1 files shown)


hlindberg:
Looks good.

There are a couple of other cases that are worth mentioning in the documentation:
* what to do if 'generate types' fails (it can in exceptional cases)
* how to opt out if there are errors (different for PE and FOSS)

jbondpdx:
Thanks @hlindberg ! I have opt-out information in PE docs, so I've got that information (and can add and link from this page). For opting out in FOSS, I assume they just don't run `generate types`, but is there more? Delete .resource_types?

And what should they do in the rare case that `generate types` doesn't work?

hlindberg:
Delete the directory `.generate_types` to opt out in FOSS, for PE there is a setting.
If genrate_types does not work - there are some tips in the training doc for support that kevin gave you a link to

jbondpdx:
@hlindberg , you said the .generate_types directory in FOSS above; is it .generate_types or .resource_types (I thought it was the latter). Also, I've updated my PR and added a ## Troubleshooting section. Would you be kind enough to give it another look?

source/puppet/4.8/env_isolation.md
```diff
@@ -0,0 +1,94 @@
...
+layout: default
+title: "Isolating resources in multiple environments"
+canonical: "/puppet/latest/environment_isolation.html"
+description: "Generating metadata to isolate resources in environments in Puppet"
+---
+
+[code_mgr_env]: {{pe}}/code_mgr.html#environment-isolation-metadata
+
+
+If you have multiple environments, environment isolation prevents resource types from leaking between your various environments. To use environment isolation, you'll generate metadata files that Puppet can use in place of the normal Ruby resource type implementations.
```
KevinCorcoran: Instead of simply "resource types", you might want to say "custom resource types" or "resource types from a module" or similar, to make it clear that we're not talking about the built-in resource types.

source/puppet/4.8/env_isolation.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Preventing resource types leaks in multiple environments
+
+If you use multiple environments with Puppet, you might encounter issues with multiple versions of the same resource type leaking between your various environments. 
+
+This problem occurs because Ruby resource type bindings are global in the Ruby runtime. Thus, the first loaded version of a Ruby resource type takes priority, and then subsequent requests to compile in other environments get that first loaded version.
+
+Environment isolation solves this issue by generating and using metadata that describes the resource type implementation, instead of using the Ruby resource type implementation when compiling catalogs. 
+
+> **Note**: Other environment isolation cases, such as external helper logic issues or varying versions of required gems, are not solved with the `generate type` command.
```
KevinCorcoran: `generate typeS` (plural)

source/puppet/4.8/env_isolation.md
KevinCorcoran: This comment might apply to a few other places in this doc as well (like the next section).

source/puppet/4.8/env_isolation.md
```diff
@@ -0,0 +1,94 @@
...
+
+1. On the command line, run `puppet generate types --environment <envname>` for each of your environments.
+
+For example, to generate metadata for your production environment, run:
+
+`puppet generate types --environment production`
+
+Whenever you deploy a new version of Puppet, overwrite previously generated metadata with `puppet generate types --environment <envname> --force`.
+
+### Implement environment isolation with r10k
```
KevinCorcoran: Sorry if this is 🔨  a 💀  🐴  but this section seems really weird after the disclaimer above: "Environment isolation is not supported for r10k with Puppet Enterprise."

source/puppet/4.8/env_isolation.md
```diff
@@ -0,0 +1,94 @@
...
+
+     1. Remove the `generate types` command from any r10k hooks.
+     1. Remove the `.resource_types` directory.
+
+  * To disable environment isolation in Puppet Enterprise:
+
+     1. In `/etc/puppetlabs/puppetserver/conf.d/pe-puppet-server.conf`, remove the `pre-commit-hook-commands` setting.
+     1. In [Hiera](./config_intro.html#configure-settings-with-hiera), set `puppet_enterprise::master::puppetserver::pre_commit_hook_commands: []`. 
+     1. On the command line, run `service pe-puppetserver reload`.
+     1. Delete the `.resource_types` directories from your staging code directory (`/etc/puppetlabs/code-staging`).
```
KevinCorcoran: Should also deploy the environment(s) after doing this so that the "un-isolated" code goes into effect (is made live).

source/puppet/4.8/env_isolation.md
hlindberg: 'non built-in resource types' ?

hlindberg:
@jbondpdx doh, {{.resource_types}} it is (edited my earlier comment)

source/puppet/4.8/env_isolation.md
jbondpdx: @KevinCorcoran and @hlindberg : In the short description at the top (the graf Kevin quoted), I can say "module resource types," if that is accurate. (I'm not sure how descriptive "custom resource types" is for less experienced users.) In the longer description of the problem, below that, I can include a statement: "This problem occurs with custom resource types in modules, not with Puppet's built-in resource types."

Does that seem accurate and reasonable? I don't know if custom types are used only within modules...?

source/puppet/4.8/env_isolation.md
jbondpdx: Yeah, this is a little tough. This is a subsection in a section about open source Puppet, on a page in the open source docs. That said, I can't say I wasn't worried about this myself.

I'll add a (re-)statement to that first graf that restates that PE + r10k isn't supported.

Otoh, if your assertion is that it's weird that we are you can use  r10k with open source Puppet but not with PE... well, you're not wrong, but I can't fix it. :-)

jbondpdx:
OK, pushed some new changes up, lemme know if they seem OK.

source/puppet/4.8/env_isolation.md
```diff
@@ -8,25 +8,25 @@ description: "Generating metadata to isolate resources in environments in Puppet
 [code_mgr_env]: {{pe}}/code_mgr.html#environment-isolation-metadata
 
 
-If you have multiple environments, environment isolation prevents resource types from leaking between your various environments. To use environment isolation, you'll generate metadata files that Puppet can use in place of the normal Ruby resource type implementations.
+If you have multiple environments, environment isolation prevents module resource types from leaking between your various environments. To use environment isolation, you'll generate metadata files that Puppet can use in place of the normal Ruby resource type implementations.
```
hlindberg: I am not so keen on the term "module resource types".

source/puppet/4.8/env_isolation.md
```diff
@@ -8,25 +8,25 @@ description: "Generating metadata to isolate resources in environments in Puppet
...
 
 In open source Puppet, enable environment isolation with the `generate types` command. In Puppet Enterprise, environment isolation is automatically provided by [Code Manager][code_mgr_env]. Environment isolation is not supported for r10k with Puppet Enterprise.
 
 ## Preventing resource types leaks in multiple environments
 
-If you use multiple environments with Puppet, you might encounter issues with multiple versions of the same resource type leaking between your various environments. 
+If you use multiple environments with Puppet, you might encounter issues with multiple versions of the same resource type leaking between your various environments. This happens with custom resource types in modules, not with Puppet's built-in resource types.
```
hlindberg: "custom resource types in modules" as opposed to what other resource types? Non custom not in modules, custom not in modules  what if they are in a gem? It can happen for all but the resource types in puppet itself.

hlindberg: Yes!

source/puppet/4.8/env_isolation.md
jbondpdx: Yeah, I can't say I love it myself. In this paragraph, I need something brief that gives the user an idea of whether this problem is of interest to them. Another option is "custom," but I think that's not great for exactly the reasons you stated.

My inclination at this point is to just leave it as "resource type," and let the more detailed conceptual section expand on that below (not Puppet built-in types, but anything else).

jbondpdx:
Cool, thanks @hlindberg and @KevinCorcoran . I'm closing this PR because I've realized I need to rename the file. Will open a clean PR for docs merge.

source/puppet/4.8/env_isolation.md
KevinCorcoran: You're not wrong, Walter ...

I guess my reaction when I saw this in the doc was "could this subsection be omitted altogether?"  But, now that I've typed that, I believe I more-or-less asked the same question when we last spoke about this in-person, so that definitely seems like 🔨  a 💀  🐴  territory.

PR #712, opened 2016-12-23
(docs) Remove version_range in sample outputs

juniorsysadmin:
Puppet 3.6.0 introduced an incorrect key name version_range when doing puppet module generate, rather than version_requirement.

This appears to have been fixed in Puppet 3.7.0 (PUP-2781)

### DIFF
--- a/source/puppet/3.7/modules_fundamentals.markdown
+++ b/source/puppet/3.7/modules_fundamentals.markdown
@@ -210,7 +210,7 @@ Where can others go to file issues about this module?
   "dependencies": [
     {
       "name": "puppetlabs-stdlib",
-      "version_range": ">= 1.0.0"
+      "version_requirement": ">= 1.0.0"
     }
   ]
 }
--- a/source/puppet/3.8/modules_fundamentals.markdown
+++ b/source/puppet/3.8/modules_fundamentals.markdown
@@ -210,7 +210,7 @@ Where can others go to file issues about this module?
   "dependencies": [
     {
       "name": "puppetlabs-stdlib",
-      "version_range": ">= 1.0.0"
+      "version_requirement": ">= 1.0.0"
     }
   ]
 }
--- a/source/puppet/4.1/modules_fundamentals.markdown
+++ b/source/puppet/4.1/modules_fundamentals.markdown
@@ -209,7 +209,7 @@ Where can others go to file issues about this module?
   "dependencies": [
     {
       "name": "puppetlabs-stdlib",
-      "version_range": ">= 1.0.0"
+      "version_requirement": ">= 1.0.0"
     }
   ]
 }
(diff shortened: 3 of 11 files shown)


gguillotte:
I'm :+1: for the issue addressed by this change, but I wouldn't mind someone with more experience with Puppet 3 taking a look at the 3.x changes to confirm it.

Running the module tool through these steps on recent versions of Puppet generates different output in a number of ways anyway, so a broader revision might be better. A quick run on 4.5+, for instance, will generate and output `myexample/examples` instead of the deprecated `myexample/tests` in the module, as noted in the doc but not demonstrated in the output. It also adds a Gemfile to the module not shown in the output, the JSON output is also formatted slightly differently, and we should swap `forge.puppetlabs.com` for `forge.puppet.com`.

For instance, here's output from 4.7:

```
$ puppet module generate examplecorp-mymodule
We need to create a metadata.json file for this module.  Please answer the
following questions; if the question is not applicable to this module, feel free
to leave it blank.

Puppet uses Semantic Versioning (semver.org) to version modules.
What version is this module?  [0.1.0]
--> 0.1.0

Who wrote this module?  [examplecorp]
--> Pat

What license does this module code fall under?  [Apache-2.0]
--> Apache 2.0

How would you describe this module in a single sentence?
--> It examples with Puppet.

Where is this module's source code repository?
--> https://github.com/examplecorp/examplecorp-mymodule

Where can others go to learn more about this module?  [https://github.com/examplecorp/examplecorp-mymodule]
--> https://forge.puppet.com/examplecorp/mymodule

Where can others go to file issues about this module?  [https://github.com/examplecorp/examplecorp-mymodule/issues]
-->

----------------------------------------
{
  "name": "examplecorp-mymodule",
  "version": "0.1.0",
  "author": "Pat",
  "summary": "It examples with Puppet.",
  "license": "Apache 2.0",
  "source": "https://github.com/examplecorp/examplecorp-mymodule",
  "project_page": "https://forge.puppet.com/examplecorp/mymodule",
  "issues_url": "https://github.com/examplecorp/examplecorp-mymodule/issues",
  "dependencies": [
    {"name":"puppetlabs-stdlib","version_requirement":">= 1.0.0"}
  ],
  "data_provider": null
}
----------------------------------------

About to generate this metadata; continue? [n/Y]
--> Y

Notice: Generating module at /Users/Pat/Development/mymodule...
Notice: Populating templates...
Finished; module generated in mymodule.
mymodule/examples
mymodule/examples/init.pp
mymodule/Gemfile
mymodule/manifests
mymodule/manifests/init.pp
mymodule/metadata.json
mymodule/Rakefile
mymodule/README.md
mymodule/spec
mymodule/spec/classes
mymodule/spec/classes/init_spec.rb
mymodule/spec/spec_helper.rb
mymodule/tests
mymodule/tests/init.pp
```

juniorsysadmin:
I would say exact output is less important than removing the mention of ```version_range```, which continues to show up and cause troubles where it is used

gguillotte:
@juniorsysadmin I don't disagree, I'm just saying we can (and should) do both.

ping @jtappa for a second set of eyes

gguillotte:
Merged this, and we can follow up with broader updates to the example output in docs. Thanks, @juniorsysadmin !

PR #713, opened 2017-01-10
(docs) Whitespace-related style rules switch from MUST to SHOULD

hubertgrzeskowiak:
Tabs vs spaces is an old war and both sides have their valid points. Common consensus is to decide this company-wide and stick to the choice. Only tabs and spaces mixed make serious problems. Regarding trailing spaces: they do not impair readability, nor do they increase the bug potential, so why burn them with fire? They do no good, but also no harm, therefore MUST -> SHOULD.

### DIFF
--- a/source/guides/style_guide.markdown
+++ b/source/guides/style_guide.markdown
@@ -8,7 +8,7 @@ The Puppet Language Style Guide
 
 #### Metadata
 
-Puppet Language Style Guide: Version 2.0.1
+Puppet Language Style Guide: Version 2.0.2
 
 Puppet: Version 3.7+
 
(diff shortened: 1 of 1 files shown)


hubertgrzeskowiak:
In case this is considered for style guide 3.0, I can make a merge request against that branch.

jbondpdx:
Hi @hubertgrzeskowiak ! There's currently a PR for a revised style guide for Puppet 4, so we can't merge this change at this point. (Once the new style guide is merged, I'll send an email to community to request feedback. That will be this week, I expect.)

To make this kind of change, we'll need a broader consensus that this really should change from a "must" to a "should." To that end, would you be willing to file a ticket about this? If you could file it as a DOCUMENT ticket and assign it to me (Jean Bond), it would be really helpful.

hubertgrzeskowiak:
Thanks for the response. Ticket to be found at https://tickets.puppetlabs.com/browse/DOCUMENT-638.

jbondpdx:
Thank you so much! I'll go ahead and close this PR.

PR #714, opened 2017-01-10
moving rev style guide into versioned docs

jbondpdx:
Revised style guide in 4.7 and 4.8; added style guide to Puppet ToC for 4.7 and 4.8; removed style guide from guides directory completely. We'll need to do a redirect for the old style guide link.


nfagerlund:
👍 ship it

PR #718, opened 2017-01-30
Update top scope variable to use facts

dan-wittenberg:


jbondpdx:
Thank you @dan-wittenberg !

PR #720, opened 2017-02-06
(DOC-3095) Update/correct example - Sensitive

carthik:
The fix ensures that the code sample shown on the doc page is an
example, however simple, of how to use encrypt the value before storing
it in a variable, using the Sensitive type.

This does not seek to make the example or the doc page perfect, just
better.


gguillotte:
:+1: Would this apply to each version back to 4.6?

carthik:
@gguillotte - yes, I assumed there would be some way to back-port the fix. Should I add the other versions to the PR?

jtappa:
@carthik I will follow up with bringing this info back to 4.6. Thanks

PR #723, opened 2017-02-23
Add rule about class/defines indentation and alignment

binford2k:


jbondpdx:
@DavidS , @hlindberg , @rnelson0 , could you take a look at this and let me know if it looks all right to you?

hlindberg:
if there is only one parameter, the rule is overkill
we do not use "tab stops" (it should be two spaces indentation)
our recommendation is to type all parameters - the examples should be consistent and use typed parameters

My personal opinion is that readability trumps the rule of breaking up parameters on multiple lines - if there is only one or two parameters, and they do not have long complicated names or types for example `define foo(Integer $x, Integer $y) { ... }` the multiline rule make the readability worse. I would therefore like the "must" to be expressed in a softer manner. My rationale for that is that some take these rules as absolute truths (and implement lint rules for them) which then pester everyone with worse and unnecessary formatting.

The same applies for the alignment rule, again it is a matter of readability. The lines can be of very different lengths due to types and the names of the variables. It is best to cluster them such that all within a ~20 char spread of each other are aligned the same way.

binford2k:
We do use stops in this guide, the language was borrowed from https://github.com/puppetlabs/puppet-docs/blob/master/source/puppet/4.9/style_guide.markdown#98-resource-attribute-indentation-and-alignment.

Adding data types is great and I'd totally vote for it to be added in its own rule.

jbondpdx:
I find two other incidents of "one stop" instead of "two spaces." I'm going to merge this and change all instances of "one stop" to "two spaces" after merge.

PR #725, opened 2017-02-23
Add note about psuedo namevar

binford2k:
This was briefly debated in the `puppet-lint` triage today. We did not
come to a consensus. Don't merge without other approval.

### DIFF
--- a/source/puppet/4.9/style_guide.markdown
+++ b/source/puppet/4.9/style_guide.markdown
@@ -857,6 +857,8 @@ class apache {
 
 In parameterized class and defined type declarations, required parameters must be listed before optional parameters (that is, parameters with defaults). Required parameters are parameters which are not set to anything, including undef. For example, parameters such as passwords or IP addresses might not have reasonable default values.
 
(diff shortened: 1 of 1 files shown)


binford2k:
Looks like you can only request a review from members of the org. @rodjek and @rnelson0, do you have opinions on this?

rodjek:
After thinking about this further, I'm generally against this exception. Even though the parameter is defaulting to a required value, the parameter itself is still optional as you don't have to manually specify it in your manifest/hieradata/etc.

rodjek:
> Looks like you can only request a review from members of the org.

I should be an org member now ;)

rnelson0:
I, too, am against an exception. It may be worth spelling out the use case AND noting that it goes with the other optionals, though.

rodjek:
Seems like we have a unanimous decision here, so I'm going to close the puppet-lint issue that triggered this.

I agree with @rnelson0 that we should probably call out in the style guide that using a required value like `$name` or `$title` as the default value for a parameter doesn't make that parameter required. @binford2k seeing as you opened the PR, would you mind updating your change to reflect this? :)

PR #726, opened 2017-03-01
(FM-5790) updates to BGTM

jbondpdx:


source/guides/module_guides/bgtm.md
```diff
@@ -58,194 +60,310 @@ In terms of class structure we recommend the following (more detail below):
...
-
-For a real world example of anchoring in action, please see Hunter's [puppet-wordpress](https://github.com/hunner/puppet-wordpress/blob/master/manifests/init.pp) module.
+Containment is supported in Puppet 3.4 and later. To support versions prior to Puppet 3.4 (or Puppet Enterprise 3.2), you must use the [anchor pattern][anchor] to hold those classes in place. Anchoring requires [puppetlabs-stdlib](http://forge.puppetlabs.com/puppetlabs/stdlib).
 
 ### 2d: Dependencies
 
-If your module's functionality depends on another module, then you must list these dependencies and include them directly, rather than indirectly hoping they are included in the catalog.
+If your module's functionality depends on another module, then you must list these dependencies and include them directly.
 
 This means you must `include x` in your `init.pp` to ensure the dependency is included in the catalog. You must also add the dependency to the module's [metadata.json](/guides/style_guide.html#module-metadata) and `.fixtures.yml`. (`.fixtures.yml` is a file used exclusively by rspec to pull in dependencies required to successfully run unit tests.)
```
DavidS: Minor nit: it should be included in the *main class*, which is a more specific location within init.pp.

source/guides/module_guides/bgtm.md
```diff
@@ -255,59 +373,59 @@ Congratulations! You have written a module that accomplishes a task; has appropr
...
-
-    require 'puppetlabs_spec_helper/rake_tasks'
 
 ## Step Four: Module versioning
 
-Modules, like any other piece of software, must be versioned and released when changes are made. We use and recommend using [SemVer 1.0.0](http://semver.org/spec/v1.0.0.html). It sets out the exact rules around when to increment major versions and so forth.
+Modules, like any other piece of software, must be versioned and released when changes are made. We use and recommend using [SemVer 1.0.0](http://semver.org/spec/v1.0.0.html). It sets out specific rules for when to increment major and minor versions.
 
-Once you've decided on the new version number, you must increase the version number metadata in the Modulefile. Versioning within the Modulefile allows you to build Puppet environments by picking and choosing specific versions of a module or set of modules.
```
DavidS: "version number metadata" should be just "version number". The second sentence can be dropped completely, or replaced by a pointing to general packaging practices (outside of puppet).

source/guides/module_guides/bgtm.md
```diff
@@ -58,256 +60,323 @@ In terms of class structure we recommend the following (more detail below):
...
+}
+```
 
 #### `module::config`
 
 The resources related to configuring the installed software should be placed in a config class. The config class must be named `module::config` and must be located in the `config.pp` file.
 
-See, for example, the ntp module:
+For example, see the `module::config` class in the `ntp` module: TODO: again, do we need this whole class here?
+
```
eputnam: TODO: should this TODO be in here? also, i think perhaps it's not necessary to include the entire class here. it's really just the show the name and maybe 10 or 12 lines, right?

source/guides/module_guides/bgtm.md
```diff
@@ -58,256 +60,323 @@ In terms of class structure we recommend the following (more detail below):
...
-~~~
+  }
+
+}
+```
 
 ### 2b: Parameters
 
-Parameters form the public API of your module. They are the most important interface you expose, and care should be taken to give a good balance to the number and variety of parameters so users can customize their interactions with the module. Below, we will walk through best practices for naming and developing parameters.
+Parameters form the public API of your module. They are the most important interface you expose, and you should take to balance to the number and variety of parameters so that users can customize their interactions with the module. Below, we walk through best practices for naming and developing parameters.
```
eputnam: "and you should take _care_ to balance"?

source/guides/module_guides/bgtm.md
```diff
@@ -58,256 +60,323 @@ In terms of class structure we recommend the following (more detail below):
...
 
-Adding parameters that allow you to override templates is an anti-pattern to avoid. Parameters that allow overriding templates lead to end users overriding your template with a custom template that contains hardcoded additional parameters. Hardcoding parameters in a template should be avoided, as it leads to stagnation and inhibits flexibility over time. It is far better to create more parameters and modify the original template, or have a parameter which accepts an arbitrary chunk of text added to the template, than it is to override the template with a customized one.
+Avoid adding parameters that allow you to override templates. When your parameters allow  template overrides, users may override your template with a custom template that contains additional hardcoded parameters. Hardcoded parameters in templates inhibits flexibility over time. It is far better to create more parameters and then modify the original template, or have a parameter which accepts an arbitrary chunk of text added to the template, than it is to override the template with a customized one.
```
eputnam: should the above section be folded into the parameters section?

source/guides/module_guides/bgtm.md
```diff
@@ -58,256 +60,323 @@ In terms of class structure we recommend the following (more detail below):
...
+contain ntp::install
+  contain ntp::config
+  contain ntp::service
+  Class['::ntp::install'] ->
+  Class['::ntp::config'] ~>
+  Class['::ntp::service']
+```
 
-For a real world example of anchoring in action, please see Hunter's [puppet-wordpress](https://github.com/hunner/puppet-wordpress/blob/master/manifests/init.pp) module.
+Containment is supported in Puppet 3.4 and later. To support versions prior to Puppet 3.4 (or Puppet Enterprise 3.2), you must use the [anchor pattern][anchor] to hold those classes in place. Anchoring requires [puppetlabs-stdlib](http://forge.puppetlabs.com/puppetlabs/stdlib).
```
eputnam: It seems like we shouldn't even mention anchoring because Puppet 3 is now EOL and if we're not going to include an example.

jbondpdx: There were discussions elsewhere (of course I can't find/remember where) that we want to de-emphasize anchoring, but we need to at least mention it.  The reality is that people still have to maintain old stuff. Hopefully beginners won't use it, but they may need to know where to find it.

source/guides/module_guides/bgtm.md
jbondpdx: This is under a subhead of Parameters, in the section Numbers. (I know that headers are harder to follow in markdown because the visual cues are missing.) So it's being treated as relating to the number of parameters; if this isn't accurate, of course, I'm happy to move it!

source/guides/module_guides/bgtm.md
jbondpdx: omg yes

source/guides/module_guides/bgtm.md
jbondpdx: haha, no TODO should never be there... I should have searched for this.

I've cut down the sample now... let me know if I cut anything especially illustrative or anything!

source/guides/module_guides/bgtm.md
eputnam: ah ok, my bad!

source/guides/module_guides/bgtm.md
eputnam: yeah totally, this makes sense.

jbondpdx:
OK, I think this is ready for merge.

- BGTM is edited and updated
- BGTM is moved out of guides and into version docs as of 4.9
- the ToC is updated
- the image from guides/module_guides has been moved to puppet-docs/puppet/source/4.9/images
- there's a redirect from guides/module_guide/bgtm.html to /puppet/latest/bgtm.html

PR #727, opened 2017-03-03
Correct the defined type signature example

binford2k:


jtappa:
This is a generated file, this is not the location to edit this content.

MFredette:
@jtappa - what's the appropriate way to make this change? Thanks!

jtappa:
Sorry, this change has to be addressed in the puppet repo

PR #729, opened 2017-03-16
(DOC-3133) Update Supported External CA Documentation

ipcrm:
This PR replaces all of the 4.x documentation for External CA support.  Prior to this PR all of the documented CA strategies applied to the older (3.x and older) RACK based puppet master configuration.

### DIFF
--- a/source/puppet/4.1/config_ssl_external_ca.markdown
+++ b/source/puppet/4.1/config_ssl_external_ca.markdown
@@ -1,88 +1,41 @@
 ---
 layout: default
-title: "SSL Configuration: External CA Support"
-canonical: "/puppet/latest/reference/config_ssl_external_ca.html"
+title: "SSL configuration: External CA support"
 ---
 
 [conf]: ./config_file_main.html
 [verify_header]: ./configuration.html#sslclientverifyheader
 [client_header]: ./configuration.html#sslclientheader
 [ca_auth]: ./configuration.html#sslclientcaauth
-[puppetdb]: /puppetdb/latest
+[puppetdb]: {{puppetdb}}/
 
 In lieu of its built-in certificate authority (CA) and public key infrastructure (PKI) tools, Puppet can use an existing external CA for all of its secure socket layer (SSL) communications.
 
-This page describes the supported and tested configurations for external CAs in this version of Puppet. If you have an external CA use case that isn't covered here, please contact Puppet Labs so we can learn more about it.
+This page describes the supported and tested configurations for external CAs in this version of Puppet. If you have an external CA use case that isn't covered here, please contact Puppet so we can learn more about it.
 
 > **Note:** This page uses RFC 2119 style semantics for MUST, SHOULD, and MAY.
 
-## Supported External CA Configurations
+## Supported external CA configurations
 
 This version of Puppet supports _some_ external CA configurations, but not every possible arrangement. We fully support the following setups:
 
 1. [Single self-signed CA which directly issues SSL certificates.](#option-1-single-ca)
-2. [Single, intermediate CA issued by a root self-signed CA.](#option-2-single-intermediate-ca) The intermediate
-   CA directly issues SSL certificates; the root CA doesn't.
-3. [Two intermediate CAs, both issued by the same root self-signed CA.](#option-3-two-intermediate-cas-issued-by-one-root-ca)
-    * One intermediate CA issues SSL certificates for Puppet master servers.
-    * The other intermediate CA issues SSL certificates for agent nodes.
-    * Agent certificates can't act as servers, and master certificates can't act as clients.
+2. [Puppet master functioning as an intermediate CA of a root self-signed CA.](#option-2-intermediate-ca)
 
-These are fully supported by Puppet Labs, which means:
+These are fully supported by Puppet, which means:
 
 * Issues that arise in one of these three arrangements are considered **bugs,** and we'll fix them ASAP.
 * Issues that arise in any _other_ external CA setup are considered **feature requests,** and we'll consider whether to expand our support.
 
-These configurations are all-or-nothing rather than mix-and-match. When using an external CA, the built-in Puppet CA service **must** be disabled and cannot be used to issue SSL certificates.
+## General notes and requirements
 
-Additionally, Puppet cannot automatically distribute certificates in these configurations --- you must have your own complete system for issuing and distributing certificates.
-
-## General Notes and Requirements
-
-### Rack Webserver is Required
-
-The Puppet master must be running inside of a Rack-enabled web server, not the built-in Webrick server.
-
-In practice, this means Apache or Nginx. We fully support any web server that can:
-
-* Terminate SSL
-* Verify the authenticity of a client SSL certificate
-* Set two request headers for:
-    * Whether the client was verified
-    * The client's distinguished name
-
-### PEM Encoding of Credentials is Mandatory
+### PEM encoding of credentials is mandatory
 
 Puppet always expects its SSL credentials to be in `.pem` format.
 
-### Normal Puppet Master Certificate Requirements Still Apply
+### Normal Puppet master certificate requirements still apply
 
 Any Puppet master certificate must contain the DNS name at which agent nodes will attempt to contact that master, either as the subject common name (CN) or as a Subject Alternative Name (DNS).
-
-### Format of X-Client-DN Request Header
-
-Rack web servers must set a client request header, which the Puppet master will check based on the [`ssl_client_header` setting][client_header].
-
-This header should conform to the following specifications:
-
-* The value of the client certificate's distinguished name (DN) should be in [RFC-2253](http://www.ietf.org/rfc/rfc2253.txt) format. The format of the `SSL_CLIENT_S_DN` environment variable (set by `mod_ssl` in Apache 2.2 and newer) is fully supported.
-* Alternatively, the value of this request header may be in "OpenSSL" format.
-
-### Revocation
-
-Certificate revocation list (CRL) checking works in all three supported configurations as long as the CRL file is distributed to the agents and masters using an "out of band" process. Puppet won't automatically update the CRL on any of the system's components.
-
-#### If Unused:
-
-If revocation lists are **not** being used by the external CA, you must disable CRL checking on the agent. Set `certificate_revocation = false` in the `[agent]` section of [puppet.conf][conf] on **every agent node.**
-
-(If it's not set to false and the agent doesn't already have a CRL file, it will try to download one from the master. This will fail, because the master must have the CA service disabled.)
-
-#### If Used:
-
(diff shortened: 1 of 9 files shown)


ahpook:
I've reviewed this and am :+1:.

@adrienthebo could you please give this doc update a technical review and :+1: if it looks good? 

@kayayarai - could you please give it a Puppet Style review? Specifically, I noticed that @ipcrm changed the markdown code offset characters but am not sure whether the tildes are still necessary. (The initial doc was several years old and hasn't been updated in a while).

kayayarai:
@ahpook @ipcrm If you want a full style review on this kind of volume, please open a separate ticket (Doc-Internal) and assign it to me; don't hold up this pull request for it.

Regarding codeblocks, the tildes and the backticks should work the same (except if you're indicating it's xml or html: then you must use ~~~). There's no reason to to change them (or not change them). Although sometimes Kramdown is weird, and you should definitely spot-check the build result to make sure it worked as desired (esp if you're putting them inside lists).

nfagerlund:
I'm taking a look at this, and I'd like to just slipstream in the info from https://docs.puppet.com/puppetserver/2.7/external_ca_configuration.html instead of linking out to it. I'll do an edit for that, and check the markdown syntax while I'm in there. Then we can get Finch to check the technical chops.

One thing I'm noticing already is that `ca = false` in puppet.conf shouldn't be needed if you're using puppetserver.

nfagerlund:
@ipcrm Could you take a look at this branch, and merge it into yours if you're cool with the changes? https://github.com/nfagerlund/puppet-docs/tree/ipcrm-external-ca-support (I can't do a PR to the PR because github kind of gives up and cries once there's a certain number of forks.)

ipcrm:
@nfagerlund Sorry for the delay; they have been merged.

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
 
-~~~ bash
-cat agent_ca.pem root_ca.pem > ca_bundle_for_master.pem
-~~~
+In order to configure the puppet master you will need the following files placed:
 
-Put this file somewhere predictable. Puppet doesn't use it directly, but it can live alongside Puppet's copies of the certificates.
+Certificate     | Purpose                     | File Location
+----------------|--------------------------------------------
+ca_crt.pem      | Intermediate CA Certificate | /etc/puppetlabs/puppet/ssl/ca/ca_crt.pem
```
adrienthebo: These paths seem suspect to be because `/etc/puppetlabs/puppet/ssl/ca` (ostensibly) should only be present on Puppet masters that are also CAs. Since Puppetserver can't operate as a CA can these files live in `/etc/puppetlabs/puppet/ssl/*` ? For instance can we keep the certs in `/etc/puppetlabs/puppet/ssl/certs`?

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
+ca_key.pem      | Intermediate CA Key         | /etc/puppetlabs/puppet/ssl/ca/ca_key.pem
```
adrienthebo: If the Puppetserver is not operating as a signing authority why should it have a copy of the intermediate CA private key?

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
+root_crt.pem    | Root CA Certificate         | /etc/puppetlabs/puppet/ssl/ca/root_crt.pem
 
-With these files in place, the web server should be configured to:
+> Note: The root_crt.pem can actually be placed anywhere, however this doc assumes you placed it as shown
```
adrienthebo: I would advise noting that the name is arbitrary but that it should be kept in `$ssldir/certs`.

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
 
-* Use the Agent+Root CA bundle, the master's certificate, and the master's key.
-* Set the verification header (as specified in the master's [`ssl_client_verify_header` setting][verify_header]).
-* Set the client DN header (as specified in the master's [`ssl_client_header` setting][client_header]).
+> Note: The ca_key.pem needs to have any passphrase removed to match the expectations of the Puppet CA
 
-An example of this configuration for Apache:
+All of the files placed should have owner set to `puppet:puppet` and permissions of `0600`.
```
adrienthebo: Slightly strange wording "should have owner set to" -> "should be owned by" ?

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
-    SSLProtocol ALL -SSLv2
-    SSLCipherSuite ALL:!ADH:RC4+RSA:+HIGH:+MEDIUM:-LOW:-SSLv2:-EXP
+Next, you have to generate the CA bundle to be placed on the master as well as any agents you create.  This is achieved by combining the Root CA and Intermediate CA certificates into one PEM file.
 
-    # Replace with the value of `puppet master --configprint hostcert`
-    SSLCertificateFile "/path/to/master.pem"
-    # Replace with the value of `puppet master --configprint hostprivkey`
-    SSLCertificateKeyFile "/path/to/master.key"
+```
+cd /etc/puppetlabs/puppet/ssl/ca
```
adrienthebo: See the earlier comments about conflating `$ssldir` (`/etc/puppetlabs/puppet/ssl`) and `$cadir` (`/etc/puppetlabs/puppet/ssl/ca`).

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
+cat ca_cert.pem root_crt.pem > ../certs/ca.pem
```
adrienthebo: In this location we just switched from using the `$cadir` to the `$ssldir/certs`.

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
+You now need to generate a new certificate for the puppet master to use.  Remember to include the `dns_alt_names` that this puppet master will need to service.
 
-    # Allow only clients with a SSL certificate issued by the intermediate CA with
-    # common name "Puppet Agent CA"  Replace "Puppet Agent CA" with the CN of your
-    # Agent CA certificate.
-    <Location />
-        SSLRequire %{SSL_CLIENT_I_DN_CN} eq "Puppet Agent CA"
-    </Location>
+```
+puppet cert generate puppetserver.my.domain.net --dns_alt_names=puppetserver,puppet
```
adrienthebo: So this explains why we're installing the intermediate CA key, but if someone is using an external CA shouldn't they be using their own tooling to generate the server certificate? Doing this with Puppet will bypass the signing processes that some companies might have.

source/puppet/4.1/config_ssl_external_ca.markdown
```diff
@@ -201,230 +107,70 @@ Credential                        | File location
...
-      |                 |                |                |
-      | Master SSL Cert |                | Agent SSL Cert |
-      |                 |                |                |
-      +-----------------+                +----------------+
-
-In this configuration, Puppet agents are configured to only authenticate peer certificates issued by the Master CA. Puppet masters are configured to only authenticate peer certificates issued by the Agent CA.
+## Option 2: Puppet master functioning as an intermediate CA
 
-> **Note:** If you're using this configuration, you can't use the ActiveRecord inventory service backend with multiple Puppet master servers. Use [PuppetDB][] for the inventory service instead.
+The puppet master can operate as an intermediate CA to an external Root CA.  The puppet master cannot be an intermediate to an intermediate.  In this mode the puppet master CA is left enabled and generation of agent certificates can remain automated, however there are some limitations:
```
adrienthebo: Tiny little nit - we're inconsistent in when we capitalize the words "intermediate" and "root" when referring to CAs.

adrienthebo:
Apologies for the delay in my review; this fell off my radar. I've made a number of comments on the 4.1 version of the docs; @ipcrm are the changes for the rest of the versions more or less equivalent? I looked around and got the impression that the changes were uniform across versions.

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: @adrienthebo In this section we are configuring puppetserver to run as an intermediate of an external ROOT CA.  So in this mode, it still is a CA and uses the default paths.

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: @adrienthebo Same as above.  The master is running as an intermediate of a ROOT CA, so it does need to have a key present.

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: @adrienthebo This is the correct path.  We are about to generate the CA bundle to be used by clients.  We start by cd'ing to $cadir and then concatenate our signing cert with the root CA's cert into one bundle located here $ssldir/certs/ca.pem.  This is the same file that we need to pass out to our agents.

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: Yep. See above.

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: @adrienthebo The puppetserver is functioning as an intermediate and is responsible for signing all agent certificates, master or otherwise.

ipcrm:
@adrienthebo Yep!  All of the changes are identical.  I replaced all 4.x docs since they were all dated and pointed to RACK based configurations which are invalid.

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: Got it; will update.

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: Got it; will update

source/puppet/4.1/config_ssl_external_ca.markdown
ipcrm: Got it; will update

ipcrm:
@adrienthebo Ugh... maybe not.  Looks like changes @nfagerlund made were only to 4.9 and you did your review on 4.1.  I'll update what we've called out here, replace all 4.x files with the 4.9, and then hopefully you can take a second look.  Technical is all the same, just formatting and wording changes.

PR #730, opened 2017-03-20
Fix setting jruby-puppet.use-legacy-auth-conf value

dud225:
Hello

I noticed a small mistake in the documentation.

Excerpt from the documentation :

>use-legacy-auth-conf in the jruby-puppet section: If true, Puppet Server
>uses the Ruby authorization methods and Puppet auth.conf rule format and
>warns you that this is deprecated. If false, Puppet Server uses the new
>authorization method and HOCON auth.conf format. Default: true.


dud225:
This PR just fixes a small mistake.

gguillotte:
Thanks, @dud225 !

PR #733, opened 2017-04-03
(DOCUMENT-662) Add rack deprecation warning

binford2k:
The deprecation warning was on the top-level rack explanation page, but
not on the passenger install page. This was a problem due to deep links
that bypass the rack intro and people still are installing passenger
setups because they don't know better.


nfagerlund:
There's a markdown syntax error in here but I'll just "fix it in post"

PR #741, opened 2017-04-21
(DOC-3177) run stages are parse order dependent

binford2k:


jtappa:
Hey @hlindberg, does this look technically accurate to you? If so we can merge and bring forward to latest docs.

hlindberg:
We prefer the terms "evaluation-order dependant" and "evaluation order" over the older (and inaccurate) "parse-order dependant" and "parse order". (Technically this has nothing to do with parsing)

hlindberg:
I can not say anything about the accuracy of the change otherwise - I don't understand run-stages. (I also do not understand what the description is telling me other than I have to be careful with ordering).

hlindberg:
@branan does this make sense to you?

branan:
I'm with Henrik that we want evaluation-order in the docs, rather than parse-order. Otherwise this looks correct to me. I had to read it a couple of times to be sure it was right, though. It could probably be written more clearly. Perhaps just trim everything after "... that uses them"?

jbondpdx:
@binford2k could you update your PR, in line with what @hlindberg and @branan said?

binford2k:
@jbondpdx 👍

PR #746, opened 2017-05-09
mapped_paths improvements

jbondpdx:
Hey @nfagerlund , would you take another look at this and either give me further feedback or merge if it looks legit to you?

Thank you!

### DIFF
--- a/source/puppet/4.10/hiera_config_yaml_5.md
+++ b/source/puppet/4.10/hiera_config_yaml_5.md
@@ -158,13 +158,15 @@ Key     | Data type | Expected value
 `paths` | Array     | Any number of file paths. This acts like a sub-hierarchy: if multiple files exist, Hiera searches all of them, in the order in which they're written.
 `glob`  | String    | One shell-like glob pattern, which might match any number of files. If multiple files are found, Hiera searches all of them in alphanumerical order.
 `globs` | Array     | Any number of shell-like glob patterns. If multiple files are found, Hiera searches all of them in alphanumerical order (ignoring the order of the globs).
-`mapped_paths` | Array or Hash     | A fact that is a collection (array or hash) of values. Hiera expands these values to produce an array of paths.
+`mapped_paths` | Array or Hash     | An array or hash, along with a temporary variable name to represent each element of the array or hash, and a path using the temporary variable in interpolation expressions. See the section about [mapped paths][#mapped-paths] for details.
 
 **Explicit file extensions are required** --- use something like `common.yaml`, not just `common`. (This is a change from prior versions of hiera.yaml, which magically guessed file extensions.)
 
 File paths are relative to the `datadir`: if the full datadir is `/etc/puppetlabs/code/environments/production/data` and the file path is set to `"nodes/%{trusted.certname}.yaml"`, the full path to the file is `/etc/puppetlabs/code/environments/production/data/nodes/<NODE NAME>.yaml`. (Absolute file paths are also allowed, but are rarely practical.)
 
 Whichever approach you choose, most of your hierarchy levels should [interpolate][interpolation] variables into the path, since that's what makes Hiera useful. For more information about crafting useful hierarchies, see [How hierarchies work][hierarchy].
+
+#### Globs {:.section}
 
 Globs are implemented with [Ruby's `Dir.glob` method][dir.glob]. In short:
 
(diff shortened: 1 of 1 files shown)


source/puppet/4.10/hiera_config_yaml_5.md
```diff
@@ -158,14 +158,16 @@ Key     | Data type | Expected value
 `paths` | Array     | Any number of file paths. This acts like a sub-hierarchy: if multiple files exist, Hiera searches all of them, in the order in which they're written.
 `glob`  | String    | One shell-like glob pattern, which might match any number of files. If multiple files are found, Hiera searches all of them in alphanumerical order.
 `globs` | Array     | Any number of shell-like glob patterns. If multiple files are found, Hiera searches all of them in alphanumerical order (ignoring the order of the globs).
-`mapped_paths` | Array or Hash     | A fact that is a collection (array or hash) of values. Hiera expands these values to produce an array of paths.
+`mapped_paths` | Array or Hash     | An array or hash, along with a temporary variable name to represent each element of the array or hash, and a path using the temporary variable in interpolation expressions. See the section about [mapped paths][#mapped-paths] for details.
```
nfagerlund: Markdown syntax: if it's a direct link instead of a "reference" style link, use parentheses in the second part instead of square brackets. `[mapped paths](#mapped-paths)`

source/puppet/4.10/hiera_config_yaml_5.md
```diff
@@ -184,20 +186,55 @@ Example:
...
+- service/b/common.yaml
+- service/c/common.yaml
+```
+
+If the first variable is a hash, then the temporary variable's value is always a `[key,value]` array. When interpolating the temporary variable, use key.subkey notation to get the part you want: `tmp.0` for the key name or `tmp.1` for the value. If the value is a nested hash, also specify the key name for the values you want to retrieve, such as `tmp.1.keyname`.
+
+For example, to find the value of the `network` fact:
+
+``` yaml
+- name:
```
nfagerlund: Oh, almost missed this: We need a name for this hierarchy level! "Networks" would work fine.

nfagerlund:
...I'm trying to use this example with 4.10.0, and I'm having a real hard time getting it to work. I might be missing something.

nfagerlund:
OK, never mind, I got it working! The example appears to be good. It's just that I ran across another wack issue with the puppet lookup CLI (`--facts` cannot set values in the main $facts hash).

source/puppet/4.10/hiera_config_yaml_5.md
```diff
@@ -184,20 +186,55 @@ Example:
...
+        - "networks/.yaml"
+        - "networks/10.0.24.0.yaml"
+        - "networks/.yaml"
+        - "networks/.yaml"
+        - "networks/.yaml"
+        - "networks/127.0.0.0.yaml"
+        - "networks/.yaml"
+        - "networks/.yaml"
+        - "networks/.yaml"
+```
```
hlindberg: Hm, I have a suspicion that it does a `unique` on the result - but I could be wrong. (If it isn't, then maybe it should - ping @thallgren ?).

hlindberg: @nfagerlund Do you think it is worth explaining why the output `"networks/.yaml" is in the result? It may be hard for a reader to understand that it is because there is no value (undef) for the part `interface.1.network`, as you could perhaps believe that the example is somehow broken otherwise.

thallgren: It doesn't do `unique` and I agree, that would be an improvement. It begs the question though, what do we do with other types of locations (`paths` and `uris`)? Should all of them be unique? If so, should the user be made aware of the problem somehow? What if a user enters duplicates an entry in `paths` for instance? I'd consider that an error. If they are unique before interpolation but becomes unique once interpolated, then perhaps that should render a warning?

jbondpdx: @nfagerlund , sanity check: In this example, where there is no value for `interface.1.network`, the result is "networks/.yaml". The effective results are 10.0.24.0 and 127.0.0.0.

nfagerlund: That's probably a good one for the CLI experience group meeting, or the CLI mailing list.

nfagerlund: Yeah, I do think it's worth explaining. @jbondpdx, I think that explanation works fine. 👍

hlindberg: I added a ticket for making paths unique...

jbondpdx:
@nfagerlund fixed up the commented sections; merge if it looks good to you?

nfagerlund:
👍

PR #748, opened 2017-05-23
(PUP-2551) Update to match error message

drewablank:
(PUP-2551) Update to match error message set in pull request https://github.com/puppetlabs/puppet/pull/5902
[puppethack]

### DIFF
--- a/source/puppet/5.0/configuration.md
+++ b/source/puppet/5.0/configuration.md
@@ -318,7 +318,7 @@ A node's certname is available in Puppet manifests as `$trusted['certname']`. (S
 for more details.)
 
 * For best compatibility, you should limit the value of `certname` to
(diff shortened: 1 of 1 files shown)


jtappa:
Hi @drewablank, thanks for your contribution during #puppethack!
I'm going to close this, you have the right idea but it looks like @binford2k pointed you to the wrong location to make this change. This file is auto-generated in the docs from the puppetlabs/puppet repo, and the appropriate file and line to make a PR against is here: https://github.com/puppetlabs/puppet/blob/master/lib/puppet/defaults.rb#L681

Feel free to open a PR in that repo and start the PR title with (docs) and ping me so I can review it there. Once merged, you won't see the change live in the docs until the next time we generate the references.

 If you'd rather not make another PR, I'll follow up and make the change later. 😄

PR #755, opened 2017-06-22
DOCUMENT-692: Add another reduce example

alexharv074:
This adds an example of how to use the `reduce` function to iteratively
update a Hash of Hashes.

The example is patched into: `source/puppet/4.10/function.md` and
`source/puppet/5.0/function.md`.

### DIFF
--- a/source/puppet/4.10/function.md
+++ b/source/puppet/4.10/function.md
@@ -2492,6 +2492,34 @@ $combine = $data.reduce( [d, 4] ) |$memo, $value| {
 # At the start of the lambda's first iteration, $memo contains [d, 4] and $value
 # contains [a, 1].
 # $combine contains [dabc, 10]
+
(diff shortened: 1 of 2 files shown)


alexharv074:
See also https://github.com/puppetlabs/puppet/pull/6005

alexharv074:
@hlindberg

jtappa:
Hi @alexharv074, this file is actually generated from the puppetlabs/puppet repo, so you'll have to add it there in the corresponding function's file. Otherwise any changes you make here will be overwritten the next time the function reference is generated.

alexharv074:
@jtappa As I mentioned in the comment, I also updated the source (see https://github.com/puppetlabs/puppet/pull/6005). The instructions in https://github.com/puppetlabs/puppet-docs/blob/master/CONTRIBUTING.md say that you should sometimes update the frozen docs here in puppet-docs, and in my view, it made sense to do so.

PR #757, opened 2017-06-23
(PUP-7579) Update RegEx for Tag

Iristyle:
- Update regular expression for tags to coincide with code changes
   made in https://github.com/puppetlabs/puppet/commit/2ac4dc0

### DIFF
--- a/source/puppet/5.0/lang_reserved.markdown
+++ b/source/puppet/5.0/lang_reserved.markdown
@@ -190,7 +190,7 @@ Parameter names should match the following regular expression:
 
 Tag names should match the following regular expression:
 
-    \A[a-z0-9_][a-z0-9_:\.\-]*\Z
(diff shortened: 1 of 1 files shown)


Iristyle:
@hlindberg looks like we have a few other places that could use regex updates in this doc.

@jtappa is this the right spot for this Puppet 5 PR?

jtappa:
Yup, this is exactly it for Puppet 5 docs. Is this the first version this regex applies to? Might want to make a note so people are aware of it.

jtappa:
Hey @Iristyle, can you please answer my question above? If it is true, I can add the note after merge, I just want to know if this needs a call out to users.

Iristyle:
@jtappa this is the first version of Puppet to apply this expanded regular expression, yes.  It might be useful to call that out here *unless* our blanket statement is "you must use Puppet on this version for broad UTF-8 support" and that's covered elsewhere.

PR #758, opened 2017-06-27
Prepare for PuppetDB 5.0 release

mullr:


source/_config.yml
```diff
@@ -669,13 +670,21 @@ documents:
       commit: ce030bb
       subdirectory: documentation
 
+  /puppetdb/5.0:
+    doc: puppetdb
+    version: "5.0"
+    nav: ./_puppetdb_nav.html
```
jtappa: since you're locking latest, probably also want to add a `hide:true` in here so it doesn't show up in the version switcher until release.

mullr:
@jtappa updated with hide:true

PR #760, opened 2017-06-27
topicized style guide and bgtm in Pup 5.0

jbondpdx:


jbondpdx:
@gguillotte or @jtappa , can you merge this if it doesn't look like I've done something that will break? Should be fine, I just don't like hitting "merge" on my own PRs.

PR #767, opened 2017-07-17
Make the no email spoofing line more clear

BillWeiss:


BillWeiss:
I welcome suggestions about phrasing here. I think it's a touch awkward, not sure what would be better.

jtappa:
@ipeldridge security stuff is usually your domain, right?

PR #770, opened 2017-07-28
Replace hiera %{server_facts.environment} with %{environment}

Bouke:
When testing with `sudo puppet lookup [name] --environment acceptance --node [fqdn] --explain`, I found that `%{server_facts.environment}` wasn't working. It produced an empty replacement. When replaced with `%{environment}`, it worked. I'm on 4.10, but it could be that other versions are affected as well.


jtappa:
@hlindberg can you please comment on technical accuracy of this? I'm not sure where Nick originally got this docs "tip" from.

jbondpdx:
@hlindberg , please can you look at this?

PR #774, opened 2017-08-11
comments and amendments

DavidS:


DavidS:
@jbondpdx , this is highly likely to not be mergable as is, but I didn't know a better way to get my comments in. 

For the record: I haven't found anything I would consider a release-blocker.

source/pdk/1.0/pdk_testing.md
```diff
@@ -87,7 +88,7 @@ Before you begin, ensure that you have written unit tests for your module, unles
 pdk test unit
 ```
 
-If there are no errors, this returns successfully as `exit code 0`, with no warnings or errors on zero examples.
+If there are no errors, this returns successfully with an exit code of zero, with no warnings or errors on zero examples. [DavidS: I don't understand the second half here. This is dubiously only true iff there are no tests yet.]
```
jbondpdx: Ohh, yeah, I know what happened here. This is the result if you are testing a freshly generated module with no added code. Is it still correct to say that if there are no errors, it returns successfully with an exit code of zero?

source/pdk/1.0/pdk_install.md
```diff
@@ -68,6 +68,10 @@ Download and install the PDK package for Windows systems.
 
 With PowerShell 3.0 or later, PDK loads automatically and the `pdk` command is now available to the prompt. If you are using PDK in PowerShell 2.0, you'll need to load the module and set the correct permissions for PDK.
 
+In some Windows installations trying to run the `pdk` command will complain about execution policy restrictions. In that case set your script execution policy to at least [RemoteSigned](https://msdn.microsoft.com/en-us/powershell/reference/5.1/microsoft.powershell.security/set-executionpolicy) by running `Set-ExecutionPolicy -ExecutionPolicy RemoteSigned`. Alternatively, you can change the `Scope` parameter of the `ExecutionPolicy` to the current session only by running `Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser`
```
DavidS: That is incorrect. The default ExecutionPolicy depends on windows edition, and local settings. Besides many enterprises, and developers changing the default, for example Home versions of Windows usually have been locked down.

The error message from powershell when people run into this is already quite helpful. Having this here only helps by having the command, background, etc ready for copy&paste, when they need it.

source/pdk/1.0/pdk_testing.md
DavidS: Yes, that is true. And rspec-puppet does not give you any warnings.

source/pdk/1.0/pdk_install.md
jbondpdx: OK, thanks @DavidS . I'm actually going to move this to a troubleshooting page, then, and will refer to that from here.

jbondpdx:
Thank you again for the feedback!

PR #776, opened 2017-08-14
Line edit pass on pdk docs

kayayarai:
Biggest change is enforcing style for user input on commands, so it's clearer what's part of the command syntax, and what's a variable/name the user supplies. Puppet style is to specify user input <LIKE_THIS> so I changed it to that, except for <module_name> and <class_name> which I did <like_that> to indicate the lowercase requirement for those names. Adjust as you see fit. If this goes against all style you've been using in module developer docs, then just disregard, but I'm curious how readers to distinguish this kind of difference. Maybe examples make it clear enough.

### DIFF
--- a/source/pdk/1.0/pdk_install.md
+++ b/source/pdk/1.0/pdk_install.md
@@ -7,19 +7,16 @@ description: "Installing the Puppet Development Kit, the shortest path to develo
 
 [troubleshoot]: ./pdk_troubleshooting.html
 
-Install the Puppet Development Kit (PDK) as your first step in creating and testing Puppet modules.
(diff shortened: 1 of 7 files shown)


source/pdk/1.0/pdk_install.md
```diff
@@ -81,16 +78,16 @@ If you are using PDK behind a proxy, you must set environment variables before r
 
 You can either set these variables on the command line before each working session or add them to your system configuration, which varies depending on the operating system.
 
-On Windows systems, run:
+On Linux-based systems or OS X, run:
```
jbondpdx: Haha, I argued with the devs about this arrangement myself (I wanted it Linux first too).  I am vindicated.

PR #779, opened 2017-08-23
(docs) DOC-3240: recommend string encapsulation and add example

jbondpdx:
I'd like to get a quick tech review on this from  @hlindberg (or @thallgren ?) and then can merge.

### DIFF
--- a/source/puppet/5.0/lang_data_hash.md
+++ b/source/puppet/5.0/lang_data_hash.md
@@ -29,7 +29,7 @@ notice($result)
 
 ## Syntax
 
-Hashes are written as a pair of curly braces containing any number of key/value pairs. Keys should be quoted with single or double quotes. A key is separated from its value by a `=>` (arrow, fat comma, or hash rocket), and adjacent pairs are separated by commas. An optional trailing comma is allowed between the final value and the closing curly brace.
+Hashes are written as a pair of curly braces containing any number of key/value pairs. A key is separated from its value by a `=>` (arrow, fat comma, or hash rocket), and adjacent pairs are separated by commas. An optional trailing comma is allowed between the final value and the closing curly brace.
 
 ``` puppet
 { 'key1' => 'val1', key2 => 'val2' }
@@ -37,7 +37,7 @@ Hashes are written as a pair of curly braces containing any number of key/value 
 { 'key1' => 'val1', key2 => 'val2', }
 ```
 
-Hash keys can be any data type, but you should generally use only strings. You should not assign a hash with non-string keys to a resource attribute or class parameter, because Puppet cannot serialize non-string hash keys into the catalog.
+Hash keys can be any data type, but generally, you should use only strings. You should quote any keys that are strings. You should not assign a hash with non-string keys to a resource attribute or class parameter, because Puppet cannot serialize non-string hash keys into the catalog.
 
 
 ```
--- a/source/puppet/5.0/lang_data_hash.md
+++ b/source/puppet/5.0/lang_data_hash.md
@@ -29,17 +29,24 @@ notice($result)
 
 ## Syntax
 
-Hashes are written as a pair of curly braces containing any number of key/value pairs. A key is separated from its value by a `=>` (arrow, fat comma, or hash rocket), and adjacent pairs are separated by commas. An optional trailing comma is allowed between the final value and the closing curly brace.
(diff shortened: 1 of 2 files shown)


source/puppet/5.1/lang_data_hash.md
```diff
@@ -29,17 +29,23 @@ notice($result)
 
 ## Syntax
 
-Hashes are written as a pair of curly braces containing any number of key/value pairs. A key is separated from its value by a `=>` (arrow, fat comma, or hash rocket), and adjacent pairs are separated by commas. An optional trailing comma is allowed between the final value and the closing curly brace.
+Hashes are written as a pair of curly braces containing any number of key/value pairs. Keys should be string encapsulated. A key is separated from its value by a `=>` (arrow, fat comma, or hash rocket), and adjacent pairs are separated by commas. An optional trailing comma is allowed between the final value and the closing curly brace.
```
hlindberg: I have issues with the term "string encapsulated" instead of something like "should be quoted with single or double quote". In addition it is somewhat confusing since any type of key is allowed (if you want to you can use say an Array, and then it should not be "quoted" or turned into string form unless it is a requirement where the hash will be used that it must be a String. (Will expand on this in a general comment).

source/puppet/5.0/lang_data_hash.md
```diff
@@ -29,17 +29,24 @@ notice($result)
...
 ``` puppet
-{ key1 => 'val1', key2 => 'val2' }
+{ 'key1' => 'val1', key2 => 'val2' }
 # Equivalent:
-{ key1 => 'val1', key2 => 'val2', }
+{ 'key1' => 'val1', key2 => 'val2', }
 ```
 
-Hash keys can be any data type, but you should generally only use strings or numbers.
+Hash keys can be any data type, but you should generally only use strings or numbers. Hash values can be any data type.
```
hlindberg: should probably say "generally only use strings" - but I think an explanation is needed why that is. See separate general comment).

jbondpdx:
@hlindberg I'm sorry about the delay here. I've updated this PR; can you tell me if the new text is accurate?

> Hash keys can be any data type, but you should generally use only strings. You should not assign a hash with non-string keys to a resource attribute or class parameter, because Puppet cannot serialize non-string hash keys into the catalog.

source/puppet/5.0/lang_data_hash.md
```diff
@@ -29,17 +29,25 @@ notice($result)
 
 ## Syntax
 
-Hashes are written as a pair of curly braces containing any number of key/value pairs. A key is separated from its value by a `=>` (arrow, fat comma, or hash rocket), and adjacent pairs are separated by commas. An optional trailing comma is allowed between the final value and the closing curly brace.
+Hashes are written as a pair of curly braces containing any number of key/value pairs. Keys should be quoted with single or double quotes. A key is separated from its value by a `=>` (arrow, fat comma, or hash rocket), and adjacent pairs are separated by commas. An optional trailing comma is allowed between the final value and the closing curly brace.
```
hlindberg: Since keys of hashes can be anything, saying that they should be quoted is not correct.

source/puppet/5.0/lang_data_hash.md
```diff
@@ -29,17 +29,25 @@ notice($result)
...
 ``` puppet
-{ key1 => 'val1', key2 => 'val2' }
+{ 'key1' => 'val1', key2 => 'val2' }
 # Equivalent:
-{ key1 => 'val1', key2 => 'val2', }
+{ 'key1' => 'val1', key2 => 'val2', }
 ```
 
-Hash keys can be any data type, but you should generally only use strings or numbers.
+Hash keys can be any data type, but you should generally use only strings. You should not assign a hash with non-string keys to a resource attribute or class parameter, because Puppet cannot serialize non-string hash keys into the catalog.
```
hlindberg: Here it could be pointed out that `you should generally use only strings (i.e. values that are in single or double quotes).`

source/puppet/5.0/lang_data_hash.md
hlindberg: It is however correct to say that "It is best if keys *that are strings* are quoted"

source/puppet/5.0/lang_data_hash.md
hlindberg: But depends on how the issue above was phrased.

jbondpdx:
> It is probably best to document what is general for the Hash data type, and then for the more constrained hash kinds of types just document how they are constrained and why.

@hlindberg  I tried to paraphrase your earlier comments, but obviously it didn't work as hoped. :) So what should it say?

PR #795, opened 2017-09-22
(DOCUMENT-729) Add more static_catalogs config setting details.

gguillotte:
-   Add `static_catalogs` to the environment.conf docs from Puppet 4.4 onward.
-   Fix references to `static_catalog` sted `static_catalogs`.
-   Make case and formatting consistent across versions.


source/puppet/5.2/static_catalogs.md
```diff
@@ -118,14 +119,14 @@ The script's standard output becomes the file's `code_content`, provided the scr
...
 
 Static catalogs are produced only by Puppet Server. The Ruby Puppet master never produces static catalogs, even when served by WEBrick or Passenger.
 
 Puppet Server also won't produce static catalogs for an agent under the following circumstances:
 
 * If the Server's `code-id-command` and `code-content-command` settings and scripts are not configured, or if the `code-id-command` returns an empty string.
-* If the agent's `static_catalogs` setting is false, either globally in [`puppet.conf`][] or in the [`environment.conf`][] file for the environment under which the agent is requesting a catalog.
+* If the Server's `static_catalogs` setting is false, either globally in [`puppet.conf`][] or in the [`environment.conf`][] file for the environment under which the agent is requesting a catalog.
```
gguillotte: I'm concerned about making this change and would appreciate tech feedback as to the expected behavior of setting `static_catalogs` to false on an agent's `puppet.conf` file. Does that disable static catalog generation for that agent only? If so, we should rephrase to not suggest that disables it globally.

gguillotte:
@joshcooper @justinstoller Tagging you guys, but I'm happy with anyone on Platform reviewing this.

source/puppet/5.2/static_catalogs.md
joshcooper: IIRC, the `static_catalogs` setting can be set in 3 different scopes:

Globally: in server's puppet.conf
Per-environment: in server's environment.conf
Per-agent: in agent's puppet.conf

By default the agent will always request a static catalog, and it's up to the server to decide whether it can generate one (based on code-id-command, etc).

PR #797, opened 2017-09-26
Debian 9 is "Stretch"

hfm:
"Jessie" is Debian 8


gguillotte:
Also fixed elsewhere in the docs in #800

PR #801, opened 2017-10-03
Remove erroneous 'inherits ntp' stanzas in beginner's module guide

antaflos:
Since the ntp class contains the other ntp classes ntp::install,
ntp::config, ntp::service it is neither necessary nor useful for these
classes to inherit from the "main" ntp class.

Class inheritance is also never touched upon in the beginner's module
guide so it seems dangerous to use inheritance in an unnecessary way
and without any explanation.

The reason inheritance even made it into the guide is probably that the
puppetlabs-ntp module made that change years ago and it was never
cleaned up.


jbondpdx:
Thank you for the PR! And sorry for the delay; I've been pretty heads down pre-PuppetConf.

PR #803, opened 2017-10-10
Bolt 1.2 docs

jbondpdx:


source/pdk/1.0/release_notes.md
```diff
@@ -5,7 +5,54 @@ canonical: "/pdk/1.0/release_notes.html"
...
+
+Released 10 October 2017
+
+### New features
+
+#### Creation of tasks
+
+PDK can now create tasks in modules. Tasks allow you to perform ad hoc actions for on-demand infrastructure change. When you create a module, PDK creates an empty `./tasks` folder. When you create a task, PDK creates template files for the task (`<TASK>.sh`) and the task metadata (`<TASK.json`>. See the [Create a task](./pdk_creating_modules.html#create-a-task) topic for details.
+
+### PDK validates task metadata
```
jbondpdx: oops, I need an extra # here

source/pdk/1.0/release_notes.md
```diff
@@ -5,7 +5,54 @@ canonical: "/pdk/1.0/release_notes.html"
...
+
+This release adds more context to ambiguous help output from `pdk help`. [PDK-482](https://tickets.puppetlabs.com/browse/PDK-482)
+
+#### Module creation adds examples and files subdirectories
+
+The `pdk new module` command now creates `./examples` and `./files` subdirectories in the new module, to provide hints on what is possible and where these things should go. [PDK-479](https://tickets.puppetlabs.com/browse/PDK-479)
+
+### Bug fixes
+
+### Running PDK commands on Windows 7 failed
```
jbondpdx: and here: #

source/pdk/1.0/pdk_creating_modules.md
```diff
@@ -30,24 +30,29 @@ When you run the `pdk new module` command, the tool requests the following infor
...
 
-After you generate a module, validate and test the module _before_ you add classes or write new code in it. This allows you to verify that the module files and directories were correctly created.
+After you create a module, validate and test the module _before_ you add classes or write new code in it. This allows you to verify that the module files and directories were correctly created.
 
-When you have validated the module, create classes by running the `pdk new class` command. The new class command also creates a test template file for the class. You can then write tests in this template to validate your class's behavior.
+When you have validated the module, create classes, defined types, and tasks by running `pdk` commands.
+
+The new class and defined type commands create a test template file for the class or defined type, so you can write tests in this template to validate your class's behavior. The new task command creates task and task metadata templates.
+
+When you create a task, PDK creates a task file in shell (`<TASK>.sh`), but you can write tasks in any language the target nodes will run. For more information about tasks, see the documentation about writing tasks.
```
DavidS: 👍 could use a link to the correct docs.

Ah, I just found them below. Is this a general style thing in docs to have those links separate, instead of inline?

source/pdk/1.0/pdk_reference.md
```diff
@@ -69,6 +69,28 @@ Argument   | Description   | Values      | Default
...
+For example:
+
+```
+cd my_module
+pdk new task my_task
+```
+
+Argument   | Description   | Values      | Default
+----------------|:---------------:|:------------------:|-------------------------
+`--template-url=<GIT_URL>` | Overrides the template to use when generating this task. | A valid Git URL or path to a local template.   | Uses the template used to generate the module. If that template is not available, the default template at [puppetlabs/pdk-module-template](https://github.com/puppetlabs/pdk-module-template) is used.
```
DavidS: "not available" should say "not specified". If the template is specified, but invalid, results are ... undefined. likely a error, or a non-working result. This needs to be fixed in the other mentions of this param too.

source/pdk/1.0/pdk_troubleshooting.md
```diff
@@ -5,9 +5,18 @@ canonical: "/pdk/1.0/pdk_troubleshooting.html"
...
 ---
 
+{:.concept}
+## PDK not in shell PATH
+ 
+PDK is not automatically added to the PATH in some shells. To fix this, add the PATH to the affected shells:
+ 
+* For zsh on OS X, add the PATH by adding the line `eval '/usr/libexec/path_helper -s'` to to the zsh resource file (`~/.zshrc`).
+ 
+* For Debian, add a symlink to `/usr/local/bin` by running `sudo ln -sv /opt/puppetlabs/bin/pdk /usr/local/bin/`
```
DavidS: This has been fixed in recent packaging and can be deleted.

source/pdk/1.0/pdk_troubleshooting.md
```diff
@@ -5,9 +5,18 @@ canonical: "/pdk/1.0/pdk_troubleshooting.html"
 description: "Troubleshooting Puppet Development Kit, the shortest path to developing better Puppet code."
 ---
 
+{:.concept}
+## PDK not in shell PATH
+ 
+PDK is not automatically added to the PATH in some shells. To fix this, add the PATH to the affected shells:
```
DavidS: With this being zsh/OSX specific this could use some rewording

source/pdk/1.0/pdk_troubleshooting.md
```diff
@@ -5,9 +5,18 @@ canonical: "/pdk/1.0/pdk_troubleshooting.html"
...
+* For zsh on OS X, add the PATH by adding the line `eval '/usr/libexec/path_helper -s'` to to the zsh resource file (`~/.zshrc`).
+ 
+* For Debian, add a symlink to `/usr/local/bin` by running `sudo ln -sv /opt/puppetlabs/bin/pdk /usr/local/bin/` 
+
 
 {:.concept}
-### Windows: Execution policy restrictions
+## Windows: Execution policy restrictions
 
 In some Windows installations, trying to run the `pdk` command causes errors related to execution policy restrictions.
```
DavidS: Unrelated: don't call this "errors" say `In some Windows installations, trying to run the `pdk` command is prohibited by the default execution policy restrictions.`

source/pdk/1.0/release_notes.md
```diff
@@ -5,7 +5,54 @@ canonical: "/pdk/1.0/release_notes.html"
...
+
+PDK can now create tasks in modules. Tasks allow you to perform ad hoc actions for on-demand infrastructure change. When you create a module, PDK creates an empty `./tasks` folder. When you create a task, PDK creates template files for the task (`<TASK>.sh`) and the task metadata (`<TASK.json`>. See the [Create a task](./pdk_creating_modules.html#create-a-task) topic for details.
+
+### PDK validates task metadata
+
+When you run `pdk validate metadata`, PDK validates both module and task metadata files.
+
+### Improvements
+
+#### CLI help is improve
```
DavidS: bad english or good meme?

source/pdk/1.0/pdk_creating_modules.md
jbondpdx: It's considered a better practice, yes. Inline links can make users feel like they should drop what they're doing and go there now, and that's usually not needed. Also, this is the way that most links will be once everything is migrated to the new docs tooling, so we've been implementing it in newer docs as we go.

None of that is to say that inline links are ALWAYS bad; sometimes we will want them.

DavidS: 👍

source/pdk/1.0/pdk_troubleshooting.md
jbondpdx: Will do! I wasn't sure if it still applied to shells we might not know about.

source/pdk/1.0/release_notes.md
jbondpdx: Why not both?

jbondpdx:
@DavidS the PR is updated, if you want to check changes. I'll merge this early tomorrow morning.

PR #804, opened 2017-10-10
Rewrite /puppet/(version)/reference links.

gguillotte:
Links to Puppet docs containing /reference/ (such as `/puppet/latest/reference/something.html`) are a legacy of an older URL structure. After  the reference directory was removed and all of its contents moved to the parent directory, the old docs site redirected such links to remove /reference/ from the URL.

This PR applies the same regex replacement from the redirect to all links in all /puppet/ content, negating the need to rewrite or redirect such links.


jtappa:
This makes me nervous doing a mass find and replace of this size, but it sounds like it would speed the import process if Mikey didn't have to do this on his end.

kayayarai:
+1

jbondpdx:
Hear, hear. +1

PR #809, opened 2017-10-19
(PUP-6452) add note for regex support in defaultfor [Do not merge, targeted for 5.4.0]

MosesMendoza:
In Puppet 5.4.0, we add support for use of regexes to match a defaultfor. Add a
mention of this to the provider development doc.

Signed-off-by: Moses Mendoza <moses@puppet.com>

### DIFF
--- a/source/puppet/5.3/provider_development.markdown
+++ b/source/puppet/5.3/provider_development.markdown
@@ -186,8 +186,12 @@ method. For instance, this is the apt provider's declaration:
 
     defaultfor :operatingsystem => :debian
 
-The same fact matching functionality is used, so again case does
-not matter.
+The same fact matching functionality as confinement is used, with one addition.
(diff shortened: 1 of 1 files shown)


MosesMendoza:
This documents https://github.com/puppetlabs/puppet/pull/6292, so that PR should be merged with this one.

Also this is against the 5.3 directory, but should _not_ be merged there. This will be released (if that PR is merged) in Puppet 5.4.0, so when that directory is created, this change should be copied there. I'm opening this PR as-is in case that doesn't happen before I depart.

MosesMendoza:
Associated PR has been merged, so once the 5.4 directory is created, this is good to move there and ship with the 5.4.0 release

gguillotte:
Added to 5.4.0 PR.

PR #810, opened 2017-10-24
Add `puppet-agent` component list up to v5.3.2

sigv:
Even though we might not have the release note lists, it is still nice to at least have an overview of the versions themselves to see what components are changing between the various `puppet-agent` versions.


gguillotte:
Thanks, @sigv! These are supposed to be updated automatically, I'll look into what's happened there.

PR #814, opened 2017-11-02
Puppet Platform 5.3.3/agent 1.10.9 docs (HOLD FOR RELEASE)

gguillotte:
-   Add Puppet Platform 5.3.3 (Puppet 5.3.3, Agent 5.3.3, Facter 3.9.3) release notes.
-   Re-pin Puppet Server to the 5.1.x branch, to publish the Server 5.1.4 docs.

Please hold for packages going live. Merging this PR should result in the docs being published.


source/puppet/5.3/release_notes_agent.md
```diff
@@ -29,6 +32,16 @@ The `puppet-agent` package installs the latest version of Puppet 5.
...
 
+## Puppet agent 5.3.3
+
+Released November 6, 2017.
+
+This release of Puppet Platform contains several Puppet and Facter bug fixes.
+
+### Component updates
+
+This release includes component updates to [Puppet 5.3.3][], [Facter 3.9.3][], [MCollective 2.11.4][], pcp-client 1.5.5, and [pxp-agent][] 1.8.1.
```
gguillotte: FYI, vanagon still shows pcp-client 1.5.4 and pxp-agent 1.8.0.

source/puppet/5.3/release_notes.markdown
```diff
@@ -20,6 +20,51 @@ Read the [Puppet 5.1](/puppet/5.1/release_notes.html) and [Puppet 5.2 release no
...
 
+## Puppet 5.3.3
+
+Released November 6, 2017.
+
+This is a bug-fix release of Puppet.
+
+-   [All issues resolved in Puppet 5.3.3](https://tickets.puppetlabs.com/issues/?jql=fixVersion%20%3D%20%27PUP%205.3.3%27)
+
+### Bug fix: Puppet can install Forge modules with many available versions
```
jtappa: Starting with 5.0, I started to try and align open source release notes with PE's release notes style guide, so my suggestions are based on that, as well as ways to make these easier to write and maintain in easyDITA later on. I realize there's not an official confluence page for open source release note guide like there is for PE, I was mostly just closely mimicking that so I didn't think it warranted its own page, but if you want, I can totally create one. 

* There are 5ish h3 headings for open source: New features, Enhancements, Known issues (Which are now on their own page), Deprecations, and Bug fixes. I'd group all of these tickets you're listing under one 'Bug fixes' h3 so you're not repeating yourself with a "Bug fix:" in every heading and bump these down to h4s.
  * Additionally, you can make the short description of the Bug fixes topic something like "These are the issues resolved in this release." Then, you don't have to append repetitive statements like "This is fixed" in individual notes, it can be assumed based on the short desc.
* We also made the decision to stop including links to every individual ticket, especially the lower priority ones where we really are describing the whole gist of the issue in the release note. Since the link to all issues resolved in Product.x is available to them, readers can still get the info they want from that one link.

gguillotte: @jtappa It would be awesome if our open source release notes conventions were standardized and documented somewhere public, even if they're the same as PE's.

source/puppet/5.3/release_notes_agent.md
jtappa: From what I can tell it doesn't *look* like either of those have releases but I'm pretty sure @MikaelSmith would be the one who can confirm whether or not those two have version bumps.

MikaelSmith: I think @branan made the final decision on that. My understanding is - because there were only test changes - we decided not to include version bumps for pcp-client and pxp-agent.

branan: Right. The version bump tickets were created, but we closed them `wontfix` because there were no code changes. We need a better workflow around when something has commits vs. when there are actual code changes that we need to ship. Turns out that's actually not easy to automate :(

gguillotte:
@mwaggett I've addressed all feedback on this PR as of 5:40 p.m. Pacific on Nov. 3.

If any feedback follows this point and it isn't resolved by the time this release's packages are live, it should not block merging this PR for release. In that scenario, please advise the people who left feedback here to file a JIRA ticket, notify me (and for this release, also ping the Docs HipChat room since we'll have several people out of the office on Monday), and we can address that feedback after the release.

PR #815, opened 2017-11-07
(DOC-2571) Document type aliases.

gguillotte:


source/puppet/4.4/lang_type_aliases.md
```diff
@@ -1,82 +1,84 @@
...
+Puppet language aliases are transparent. This example's `notice` returns `true`:
 
-Store the type aliases you write in your modules' `types` folder, which is a top-level directory, a sibling of `manifests` and `lib`. Define only one alias per file.
+``` puppet
+type MyInteger = Integer
+notice MyInteger == Integer
+```
 
-The name of the `.pp` file defining the alias must be the alias name in all lower case without underscore separators inserted at camel case positions. For example, `MyType` is expected to be loaded from a file named `"mytype.pp"`.
+> **Note:** The internal types `TypeReference` and `TypeAlias` are never values in a Puppet program.
```
jtappa: Should instead be something to the effect of: "You should never use or alter `TypeReference` or `TypeAlias` in your implementation of Puppet, as they're internal types." 'Puppet program' is a meaningless henrikism

PR #817, opened 2017-11-15
Update type.md

clairecadman:
Based on information from an old ticket. Please review for accuracy. Thanks!


glennsarti:
This looks like it was autogenerated from puppet source.

https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/package/gem.rb#L5-L16

Should perhaps this fix land there?

clairecadman:
@glennsarti Yes you are right, I will make the change there. Thanks!

PR #818, opened 2017-11-21
Fix for typo on the "Creating Environments" page

paulgalow:
Just a fix for a small but persistent typo.


jbondpdx:
Thank you @paulgalow !

PR #821, opened 2017-11-22
Update hiera_config_yaml_5.md

clairecadman:
Adding a simple example of an environment Hiera 5 hiera.yaml file.
(https://tickets.puppetlabs.com/browse/DOC-3401)

### DIFF
--- a/source/puppet/5.3/hiera_config_yaml_5.md
+++ b/source/puppet/5.3/hiera_config_yaml_5.md
@@ -47,6 +47,25 @@ Related topics: [$confdir][confdir], [`<MODULE>`][module].
 -   `default_hierarchy` - An array of hashes, which sets a default hierarchy to be used only if the normal hierarchy entries do not result in a value. Only allowed in a module's hiera.yaml.
 -   `defaults` - A hash, which can set a default datadir, backend, and options for hierarchy levels.
 
+Example of an environment hiera.yaml file:
+
+```
+---
+version: 5
+
+defaults:
+  datadir: "data"
+
+hierarchy:
+  - name: 'Yaml backend'
+    data_hash: yaml_data
+    paths:
+      - "nodes/%{trusted.certname}.yaml"
+      - 'common.yaml'
+```
+
+Options for tweaking the hiera.yaml file:
+
 ``` yaml
 ---
 version: 5


source/puppet/5.3/hiera_config_yaml_5.md
```diff
@@ -47,6 +47,25 @@ Related topics: [$confdir][confdir], [`<MODULE>`][module].
...
 -   `defaults` - A hash, which can set a default datadir, backend, and options for hierarchy levels.
 
+Example of an environment hiera.yaml file:
+
+```
+---
+version: 5
+
+defaults:
+  datadir: "data"
```
hlindberg: this is the default datadir location -no need to explicitly set the default as the default.
If here to show that you can set it, it should be commented out.

source/puppet/5.3/hiera_config_yaml_5.md
```diff
@@ -47,6 +47,25 @@ Related topics: [$confdir][confdir], [`<MODULE>`][module].
...
+
+hierarchy:
+  - name: 'Yaml backend'
```
hlindberg: This is not a good name - the name is supposed to document the purpose why this level is in the hierarchy

source/puppet/5.3/hiera_config_yaml_5.md
```diff
@@ -47,6 +47,25 @@ Related topics: [$confdir][confdir], [`<MODULE>`][module].
...
+    data_hash: yaml_data
+    paths:
+      - "nodes/%{trusted.certname}.yaml"
+      - 'common.yaml'
```
hlindberg: These two entries have different purpose, one is the "node specific" entry, and one is the "common/defaults". They should really be in an entry each.

source/puppet/5.3/hiera_config_yaml_5.md
```diff
@@ -47,6 +47,25 @@ Related topics: [$confdir][confdir], [`<MODULE>`][module].
...
+```
+---
+version: 5
+
+defaults:
+  datadir: "data"
+
+hierarchy:
+  - name: 'Yaml backend'
+    data_hash: yaml_data
```
hlindberg: IIRC, this is also the default - think it can be left out - ping @thallgren.
If it is the default, it (together with the datadir) can be shown as a comment.

thallgren: Yes, `data_hash: yaml_data` and `datadir: data` are the defaults.

source/puppet/5.3/hiera_config_yaml_5.md
thallgren: I think that how hierarchy entries are divided depends on the purpose of the entries. I would for instance recommend that an entry that shares common options (an example of this is is an eyaml configuration) will benefit from not repeating such options just to separate paths, even if the paths have a different purpose.

clairecadman:
@hlindberg Thanks for your comments! Could you please provide an example you think would be more appreciate?

PR #826, opened 2017-12-13
Update hiera_config_yaml_5.md

clairecadman:


source/puppet/5.3/hiera_config_yaml_5.md
```diff
@@ -64,7 +64,9 @@ hierarchy:
       - 'common.yaml'
 ```
 
-Options for tweaking the hiera.yaml file:
+Note that `data_hash: yaml_data` and `datadir: data` are the default settings.
```
hlindberg: I find this confusing; are they like "factory default", or did the example above define them as defaults? If so, how did the use of `data_hash: yaml_data` make that default, it is not in the `defaults`. 

If those lines were commented out it would make sense.
```
--
 version: 5

# defaults:
#   datadir: "data"
 		 
hierarchy:
   - name: 'Yaml backend'
#     data_hash: yaml_data
     paths:	
       - "nodes/%{trusted.certname}.yaml"		       
        - 'common.yaml'
  ```

File: source/puppet/5.3/hiera_config_yaml_5.md [hunk 2]
```diff
@@ -64,7 +64,9 @@ hierarchy:
...
+
+A more advanced example of a hiera.yaml file:
```
hlindberg: In the example below - it make sense to (again) comment out the defaults, so that it begins with this:
```
version: 5
# defaults:  # Used for any hierarchy level that omits these keys.
#   datadir: data         # This path is relative to hiera.yaml's directory.
#   data_hash: yaml_data  # Use the built-in YAML backend.
```

PR #827, opened 2017-12-13
Update hiera_custom_backends.md

clairecadman:
This info was originally written about Puppet::LookupKey and Puppet::LookupValue.

### DIFF
--- a/source/puppet/5.3/hiera_custom_backends.md
+++ b/source/puppet/5.3/hiera_custom_backends.md
@@ -59,6 +59,10 @@ For more information, please the see data_dig backend reference.
 
 Related topics: [custom Puppet function][puppet_functions], [the modern Ruby functions API][ruby_functions].
 
+#### The `RichDataKey` and `RichData` types
+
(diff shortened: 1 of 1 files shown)


source/puppet/5.3/hiera_custom_backends.md
```diff
@@ -59,6 +59,10 @@ For more information, please the see data_dig backend reference.
 
 Related topics: [custom Puppet function][puppet_functions], [the modern Ruby functions API][ruby_functions].
 
+#### The `RichDataKey` and `RichData` types
+
+To simplify backend function signatures, you can use two extra data type aliases: `RichDataKey`, and `RichData`. These are only available to backend functions called by Hiera; normal functions and Puppet code can not use them.
+
```
hlindberg: That is not correct - it is a mix of old and new information. The data types `RichDataKey` and `RichData` are available everywhere. The types they replace (the old `Puppet::LookupKey` and `Puppet::LookupValue`) were originally restricted. Those types are now aliases to `RichDataKey` and `RichData` respectively - and those aliases are also available everywhere.

Suggest writing:

> To simplify writing hiera 5 backend function signatures, you can use the data types `RichDataKey` for the key(s), and `RichData` for the returned value for a key.

PR #828, opened 2017-12-13
(DOCUMENT-779) Remove escape sequence CRLF breaks.

gguillotte:
Remove a line from the documentation stating that the `/r/n` escape
sequence can create CRLF line breaks in template file output, per
[DOCUMENT-779](https://tickets.puppetlabs.com/browse/DOCUMENT-779)
and [PUP-8240](https://tickets.puppetlabs.com/browse/PUP-8240).


gguillotte:
This has been up for review for a week. I'll merge this change; if there are concerns, I can revert.

PR #830, opened 2017-12-18
(DOC-2699) Restore and add ppAuthCertExt OID range/certext docs

gguillotte:
At some point the ppAuthCertExt OID range was removed from the _registered_oids.md partial. Restore those and add them to all versions since 4.6. Add minimal docs about the auth range.

**Two files were changed — _registered_oids.md and ssl_attributes_extensions.markdown — but this PR modifies 21 files because the same changes to are duplicated across all relevant versions of the Puppet docs. No need to review each version unless different behavior is expected for specific versions.**

### DIFF
--- a/source/puppet/5.0/ssl_attributes_extensions.markdown
+++ b/source/puppet/5.0/ssl_attributes_extensions.markdown
@@ -111,7 +111,7 @@ Visibility of extensions is somewhat limited:
 * The `puppet cert list` command _does not_ display custom attributes for any pending CSRs, and [basic autosigning (autosign.conf)][autosign_basic] doesn't check them before signing. Either use [policy-based autosigning][autosign_policy] or inspect CSRs manually with the `openssl` command (see below).
 * The `puppet cert print` command _does_ display any extensions in a signed certificate, under the "X509v3 extensions" section.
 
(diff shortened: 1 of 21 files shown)


gguillotte:
Tagging @jessereynolds for reporting this, to confirm that this addresses the issue raised in the ticket.

Tagging @justinstoller for a sanity check on the tech side. I don't think anyone who worked on this in-house is still here.

source/puppet/5.0/ssl_attributes_extensions.markdown
```diff
@@ -109,7 +111,7 @@ Visibility of extensions is somewhat limited:
 * The `puppet cert list` command _does not_ display custom attributes for any pending CSRs, and [basic autosigning (autosign.conf)][autosign_basic] doesn't check them before signing. Either use [policy-based autosigning][autosign_policy] or inspect CSRs manually with the `openssl` command (see below).
 * The `puppet cert print` command _does_ display any extensions in a signed certificate, under the "X509v3 extensions" section.
 
-Puppet's authorization system (`auth.conf`) does not use certificate extensions.
+Puppet's authorization system (`auth.conf`) does not use certificate extensions, but [Puppet Server's authorization system](/puppetserver/2.7/config_file_auth.html), which is based on `trapperkeeper-authorization`, can use extensions in the ppAuthCertExt OID range, and requires them for requests to write access rules.
```
justinstoller: This is pointing to 2.7 docs for server for 5.0 platform.

justinstoller:
I think that the 5.x docs should point to puppetserver docs newer than 2.x? Otherwise this looks good.

PR #831, opened 2018-01-04
(DOC-3437) Edit and clarify SSL autosigning docs.

gguillotte:
Autosigning works slightly differently in open source Puppet, monolithic Puppet Enterprise, and split Puppet Enterprise installations. Clarify this, and reorganize other docs on basic autosigning for additional clarification.

### DIFF
--- a/source/puppet/5.3/ssl_autosign.markdown
+++ b/source/puppet/5.3/ssl_autosign.markdown
@@ -18,9 +18,13 @@ Alternately, you can configure the CA Puppet master to automatically sign certai
 
 ## Disabling autosigning
 
-By default, the CA Puppet master will use the `$confdir/autosign.conf` file as a whitelist; [see "Basic Autosigning" below][inpage_basic]. Since this file doesn't exist by default, autosigning is implicitly disabled, and the CA will not autosign any certificates.
+By default, the CA Puppet master's `autosign` setting defaults to `$confdir/autosign.conf`.
 
-To _explicitly_ disable autosigning, you can set `autosign = false` in the `[master]` section of the CA Puppet master's puppet.conf. This will cause the CA to never autosign even if an autosign.conf file is written later.
+In open source Puppet, this file doesn't exist by default. In Puppet Enterprise, it exists but is empty by default. In either case, the autosigning feature is enabled but the whitelist is empty, so the CA Puppet master doesn't autosign any certificates. In other words, the feature is implicity disabled until the `autosign.conf` file contains a whitelist or is a custom policy executable. For more information, see [basic autosigning][inpage_basic] and [policy-based autosigning][inpage_policy].
+
+To _explicitly_ disable autosigning, set `autosign = false` in the `[master]` section of the CA Puppet master's `puppet.conf`, which disables CA autosigning even if autosign.conf or a custom policy executable exists.
+
+For more information about the `autosign` setting, see the [configuration reference](./configuration.html#autosign).
 
 ## Naïve autosigning
 
@@ -28,13 +32,13 @@ Naïve autosigning causes the CA to autosign **all** CSRs.
 
 ### Enabling naïve autosigning
 
-To enable naïve autosigning, set `autosign = true` in the `[master]` section of the CA Puppet master's puppet.conf.
+To enable naïve autosigning, set `autosign = true` in the `[master]` section of the CA Puppet master's `puppet.conf`.
 
 ### Security implications of naïve autosigning
 
 **You should never do this in a production deployment.** Naïve autosigning is only suitable for temporary test deployments that are incapable of serving catalogs containing sensitive information.
 
-## Basic Autosigning (autosign.conf)
+## Basic autosigning (autosign.conf)
 
 [inpage_basic]: #basic-autosigning-autosignconf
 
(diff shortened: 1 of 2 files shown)


source/puppet/5.3/ssl_autosign.markdown
```diff
@@ -6,75 +6,73 @@ title: "SSL configuration: autosigning certificate requests"
...
+The `autosign.conf` whitelist file's location and contents are described in [its documentation](./conf_file_autosign.html).
 
-> **Note:** Basic autosigning is enabled by default and looks for a whitelist located at `$confdir/autosign.conf`. For more info, see [the page on the confdir](./dirs_confdir.html).
+Puppet looks for `autosign.conf` at the path configured in the [`autosign` setting][autosign setting] in the `[master]` section of `puppet.conf`. The default path is `$confdir/autosign.conf`, and the default `confdir` path depends on your operating system. [See the confdir documentation for more information.](./dirs_confdir.html)
 
-### The `autosign.conf` file
+> **Note:** In open source Puppet, no `autosign.conf` file exists by default. In Puppet Enterprise, the file exists but is empty. In both cases, basic autosigning is enabled by default but doesn't autosign any certificates because an blank or non-existent whitelist is considered empty and no hostnames are whitelisted by default.
```
haus: `exists but is empty` => `exists but may be empty`

source/puppet/5.3/ssl_autosign.markdown
```diff
@@ -90,27 +88,26 @@ If you aren't embedding additional data, the CSR will contain only the node's ce
...
-* **Exit status:** The executable must exit with a status of `0` if the certificate should be autosigned; it must exit with a non-zero status if it should not be autosigned.
-    * The Puppet master will treat all non-zero exit statuses as equivalent.
-* **Stdout and stderr:** Anything the executable emits on stdout or stderr will be copied to the Puppet master's log output at the `debug` log level. Puppet will otherwise ignore the executable's output; only the exit code is considered significant.
+-   **Run environment:** The executable will be run once for each incoming CSR.
+    -   It will be executed by the Puppet master process and will run as the same user as the Puppet master.
+    -   The Puppet master process will _block until the executable finishes running._ We expect policy executables to finish in a timely fashion; if they do not, it's possible for them to tie up all available Puppet master threads and deny service to other agent nodes. If an executable needs to perform network requests or other potentially expensive operations, the author is in charge of implementing any necessary timeouts, possibly bailing and exiting non-zero in the event of failure.
```
haus: I'm not sure if the nuance here matters, but a signing request will tie up a single jruby instance, so depending on how many jrubies are configured, this might not block all requests (also non-ruby requests will still be serviced).

source/puppet/5.3/ssl_autosign.markdown
```diff
@@ -90,27 +88,26 @@ If you aren't embedding additional data, the CSR will contain only the node's ce
...
+    -   (Note that under a Rack server like Passenger, there are generally multiple Puppet master processes available at any given time, so policy executables have a little bit of leeway.)
```
haus: Are we still documenting rack/passenger stuff?

source/puppet/5.3/ssl_autosign.markdown
```diff
@@ -6,75 +6,73 @@ title: "SSL configuration: autosigning certificate requests"
...
+> **Note:** In open source Puppet, no `autosign.conf` file exists by default. In Puppet Enterprise, the file exists but is empty. In both cases, basic autosigning is enabled by default but doesn't autosign any certificates because an blank or non-existent whitelist is considered empty and no hostnames are whitelisted by default.
 
-The `autosign.conf` whitelist file is a list of certnames or domain name globs (one per line) whose certificate requests will automatically be signed.
-
-    rebuilt.example.com
-    *.scratch.example.com
-    *.local
-
-Note that domain name globs do not function as normal globs: an asterisk can only represent one or more subdomains at the front of a certname that resembles a fully-qualified domain name. (That is, if your certnames don't look like FQDNs, you can't use `autosign.conf` to full effect.)
+The `autosign.conf` file must not be executable by the Puppet master's user account. If the `autosign` setting points to an executable file, Puppet instead treats it like a custom policy executable even if it contains a valid `autosign.conf` whitelist.
```
haus: what happens if it is executable by the Puppet master's user account?

source/puppet/5.3/ssl_autosign.markdown
gguillotte: That could be a good note for the "differences from Puppet" page on Puppet Server.

source/puppet/5.3/ssl_autosign.markdown
gguillotte: We probably shouldn't. My only edits here were to fix Markdown, I didn't review the words.

source/puppet/5.3/ssl_autosign.markdown
gguillotte: > Puppet instead treats it like a custom policy executable even if it contains a valid `autosign.conf` whitelist.

Puppet tries to run it and is probably pretty bummed by the result.

haus: I think maybe that nuance would be valuable to include. I didn't get that on my first read. Maybe
"If the `autosign` setting points to an executable file, Puppet instead treats it like a custom policy executable, and attempts to execute it, even if it contains a valid `autosign.conf` whitelist."

gguillotte:
Clarified bits on @haus's feedback.

PR #833, opened 2018-01-16
(DOC-3443)(PUP-6692) Document $settings::all_local

gguillotte:
`$settings::all_local` made the Puppet 5.0 release notes, but was not added to the documentation. Add a description to all Puppet 5.x docs.

### DIFF
--- a/source/puppet/5.0/lang_facts_and_builtin_vars.markdown
+++ b/source/puppet/5.0/lang_facts_and_builtin_vars.markdown
@@ -245,6 +245,7 @@ These are **not** available in the `$facts` hash.
 * `$serverip` --- the Puppet master's IP address.
 * `$serverversion` --- the current version of Puppet on the Puppet master.
 * `$settings::<name of setting>` (also available to `puppet apply`) --- the value of any of the master's [settings](./config_about_settings.html). This is implemented as a special namespace and these variables must be referred to by their qualified names. Note that, other than `$environment` and `$clientnoop`, the agent node's settings are **not** available in manifests. If you wish to expose them to the master in this version of Puppet, you will have to create a custom fact.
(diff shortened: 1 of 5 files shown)


source/puppet/5.0/lang_facts_and_builtin_vars.markdown
```diff
@@ -245,6 +245,7 @@ These are **not** available in the `$facts` hash.
 * `$serverip` --- the Puppet master's IP address.
 * `$serverversion` --- the current version of Puppet on the Puppet master.
 * `$settings::<name of setting>` (also available to `puppet apply`) --- the value of any of the master's [settings](./config_about_settings.html). This is implemented as a special namespace and these variables must be referred to by their qualified names. Note that, other than `$environment` and `$clientnoop`, the agent node's settings are **not** available in manifests. If you wish to expose them to the master in this version of Puppet, you will have to create a custom fact.
+* `$settings::all_local` --- contains all variables in the `$settings` namespace as a Hash of `<SETTING-NAME> => <SETTING-VALUE>`. This helps you find settings that might be missing when the `--strict_variables` setting is enabled.
```
hlindberg: Well, it isn't exactly "find settings that might be missing", as it helps you find those that are actually defined ;-)
It helps you refer to a setting that is potentially missing - i.e. `$settings::all_local[some_setting]` returns `undef` if not set where $settings::some_setting would raise an error if not set and `--strict_variables` is on. 

Maybe say something like:
"This helps you reference settings that may be missing as a direct reference to such a missing setting raises an error for users having enabled --strict_variables"

gguillotte: Thanks. I'll also update the release note, because this language was also used there.

PR #834, opened 2018-01-23
(DOC-3463) Include module namespacing in examples.

gguillotte:
> The information provided in https://puppet.com/docs/puppet/5.3/lang_type_aliases.html#creating-type-aliases does not highlight the need to namespace the type when creating or using the type. This lack of information prevents the Puppet user from understanding how to create or use Type Aliases.

Resolve this omission by adding namespacing to the examples.

### DIFF
--- a/source/puppet/4.10/lang_type_aliases.md
+++ b/source/puppet/4.10/lang_type_aliases.md
@@ -1,7 +1,6 @@
 ---
 layout: default
 title: "Language: Type Aliases"
-canonical: "/puppet/latest/lang_type_aliases.html"
 ---
 
 [reserved]: ./lang_reserved.html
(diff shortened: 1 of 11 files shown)


beergeek:
@gguillotte that is a lot of files! Looks good to me.

source/puppet/4.10/lang_type_aliases.md
```diff
@@ -13,21 +12,21 @@ Type aliases allow you to create reusable and descriptive data and resource type
 Type aliases are written as:
 
 ``` puppet
-type <ALIAS NAME> = <TYPE DEFINITION>
+type <MODULE NAME>::<ALIAS NAME> = <TYPE DEFINITION>
 ```
 
-The `<ALIAS NAME>` begins with a capital letter and must not be a [reserved word][reserved].
+The `<MODULE NAME>` must be named after the module that contains the Type Alias, and both the `<MODULE NAME>` and `<ALIAS NAME>` begin with a capital letter and must not be a [reserved word][reserved].
```
jtappa: I think type alias should be lowercase here, looks like it is in other places in the doc.

gguillotte:
Thanks @jtappa @beergeek! Merging this PR.

PR #837, opened 2018-01-25
(DOC-3462)(FACT-409) execution.exec is deprecated.

gguillotte:
Facter.execution.exec was deprecated in Facter 2.0. Remove it from the custom facts doc and Facter 2.x's fact overview, and add a note about its deprecation to the 2.0 release notes.


gguillotte:
@ahpook Commit 9da6ada has the relevant changes to the custom facts doc. The other commits make that same change in other doc versions.

PR #838, opened 2018-01-26
Restore minor upgrade docs.

gguillotte:
The minor upgrade docs were being published and linked to, but
weren't in the sidebar navigation. They also didn't note their relevance
to upgrades from Puppet 4.

Restore the docs to the nav and include the guidance about Puppet 4
upgrades from the index. Also, point to the macOS agent install docs for
upgrading those agents.

### DIFF
--- a/source/puppet/5.0/_puppet_toc.html
+++ b/source/puppet/5.0/_puppet_toc.html
@@ -44,6 +44,7 @@
         * [Upgrade agents from 3.8.x Puppet 5.x]({{puppet}}/upgrade_major_agent.html)
         * [Upgrade servers]({{puppet}}/upgrade_major_server.html)
         * [Post-upgrade cleanup]({{puppet}}/upgrade_major_post.html)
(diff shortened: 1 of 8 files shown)


gguillotte:
CC @ahpook @melamos regarding the install docs meeting today.

The index already pointed to the upgrade_minor docs for upgrading from Puppet 4 to Puppet 5 and was added a while ago, so I'm copying that guidance from the index into the upgrade doc. I'm not sure if it's still accurate.

gguillotte:
also pinging @jtappa in case there's additional context I'm missing

jtappa:
I can't for the life of me remember why this would've been removed from the nav, but I do feel like there was some discussion about renaming this page and getting rid of the major/minor distinction in pup 5 docs, since an upgrade from Puppet 4 to 5 isn't considered major, but from 3 it would be. Involved in that would've been renaming this file to just upgrade.md, but it wouldn't have meant removing it from the sidebar nav. ¯\_(ツ)_/¯ It could've just been a flub on mine or somebody else's part, we have had a lot of re-org and tweaks to the sidebar nav over the last two years.

PR #851, opened 2018-02-16
(DOCUMENT-817) Add String-to-Number conversion option using new()

gguillotte:
Authored string-to-number docs were not updated when Puppet 4.5.0 added `new()`, resulting in deprecated information remaining in Puppet 4.10 and 5.x docs.

-   Update those docs to reflect new methods of converting strings to other types.
-   Add a deprecation note for numeric coercion.
-   Add examples.

### DIFF
--- a/source/puppet/4.10/lang_data_number.md
+++ b/source/puppet/4.10/lang_data_number.md
@@ -7,7 +7,6 @@ title: "Language: Data types: Numbers"
 [data type]: ./lang_data_type.html
 [variant]: ./lang_data_abstract.html#variant
 
-
 Numbers in the Puppet language are normal integers and floating point numbers.
 
 You can work with numbers using Puppet's [arithmetic operators.][arithmetic]
@@ -16,38 +15,38 @@ You can work with numbers using Puppet's [arithmetic operators.][arithmetic]
 
 Numbers are written without quotation marks, and can consist only of:
 
-* Digits
-* An optional negative sign (`-`; this is actually [the unary negation operator](./lang_expressions.html#subtraction-and-negation) rather than part of the number)
-    * Explicit positive signs (`+`) aren't allowed, since there's no unary `+` operator.
-* An optional decimal point (which results in a floating point value)
-* A prefix, for octal or hexidecimal bases
-* An optional `e` or `E` for scientific notation of floating point values
+-   Digits.
+-   An optional negative sign (`-`). This is the [unary negation operator](./lang_expressions.html#subtraction-and-negation), rather than part of the number.
+    -   Explicit positive signs (`+`) aren't allowed, because there's no unary `+` operator.
+-   An optional decimal point, which results in a floating point value.
+-   A prefix, for octal or hexidecimal bases.
+-   An optional `e` or `E` for scientific notation of floating point values.
 
 ### Integers
 
 Integers are numbers without decimal points.
 
-If you divide two integers, the result _will not_ be a float; instead, Puppet will truncate the remainder. (That is, `2/3 == 0`.)
+If you divide two integers, the result _is not_ a floating point number. Instead, Puppet truncates the remainder. For example, `2/3 == 0`.
 
 ### Floating point numbers
 
-Floating point numbers ("floats") are numbers that include a fractional value after a decimal point (even if that fractional value is zero, like `2.0`).
+Floating point numbers ("floats") are numbers that include a fractional value after a decimal point, even if that fractional value is zero (such as `2.0`).
 
-If an expression includes both integer and float values, the result will be a float.
+If an expression includes both integer and float values, the result is a float.
 
 ``` puppet
 $some_number = 8 * -7.992           # evaluates to -63.936
 $another_number = $some_number / 4  # evaluates to -15.984
 ```
 
-Floating point numbers between -1 and 1 cannot start with a bare decimal point; they must have a zero before the decimal point.
+Floating point numbers between -1 and 1 cannot start with a bare decimal point. They must have a zero before the decimal point.
 
 ``` puppet
-$product = 8 * .12 # syntax error
-$product = 8 * 0.12 # OK
+$product = 8 * .12   # syntax error
+$product = 8 * 0.12  # OK
 ```
 
-You can express floating point numbers in scientific notation: append `e` or `E` plus an exponent, and the preceding number will be multiplied by 10 to the power of that exponent. Numbers in scientific notation are always floats.
+To express floating point numbers in scientific notation, append `e` or `E` and an exponent. Puppet multiplies the preceding number by 10 to the power of that exponent. Numbers in scientific notation are always floats.
 
 ``` puppet
 $product = 8 * 3e5  # evaluates to 240000.0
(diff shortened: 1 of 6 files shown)


gguillotte:
@thallgren @hlindberg Relevant changes are in https://github.com/puppetlabs/puppet-docs/pull/851/commits/eec0033fabb95409a612cf1cb3ab39340cfcb052#diff-b0c80cf31b92d3db142070174f496a8b.

The rest of this PR copies those changes to previous versions of Puppet docs and edits text and formatting for consistency.

source/puppet/4.10/lang_data_number.md
```diff
@@ -57,49 +56,63 @@ $product = 8 * 3e5  # evaluates to 240000.0
...
 Integer values can be expressed in decimal notation (base 10), octal notation (base 8), and hexadecimal notation (base 16).
 
-* _Non-zero_ decimal integers must not start with a `0`.
-* Octal values have a prefix of `0`, which can be followed by a sequence of octal digits 0-7.
-* Hexadecimal values have a prefix of `0x` or `0X`, which can be followed by hexadecimal digits 0-9, a-f, or A-F.
+-   _Non-zero_ decimal integers must not start with a `0`.
+-   Octal values have a prefix of `0`, which can be followed by a sequence of octal digits 0-7.
+-   Hexadecimal values have a prefix of `0x` or `0X`, which can be followed by hexadecimal digits 0-9, a-f, or A-F.
 
 Floats can't be expressed in octal or hex.
```
thallgren: This is wrong. All examples below (lines 66 to 76) are about integers. We don't allow floats expressed as octal and hex

source/puppet/4.10/lang_data_number.md
```diff
@@ -57,49 +56,63 @@ $product = 8 * 3e5  # evaluates to 240000.0
...
 ## Converting strings to numbers
 
-The [arithmetic operators][arithmetic] will automatically convert strings to numbers.
+[Arithmetic operators][arithmetic] automatically convert strings to numbers.
 
 In all other contexts (resource attributes, function arguments, etc.), Puppet _won't_ automatically convert strings to numbers, but you can:
 
-* Add 0 to manually convert a string to a number. (For example, `$mystring = "85"; $mynum = 0 + $mystring`.)
-* Use [the `scanf` function](./function.html#scanf) to manually extract numbers from strings. This function can also account for surrounding non-numerical text.
+-   Create a new `Integer` or `Float` object using the string as a parameter, such as `$mynum = Integer($mystring)`. This invokes the type's `new()` function, which can convert objects of other types --- including Strings --- into [Integers](./function.html#conversion-to-integer) and [Floats](./function.html#conversion-to-float).
```
thallgren: Perhaps explicitly mention that `Integer.new(args)` is exactly the same as `Integer(args)` or more generically that `T(args)` where `T` is a type name is always interpreted as `T.new(args)`?

source/puppet/4.10/lang_data_number.md
```diff
@@ -16,38 +15,38 @@ You can work with numbers using Puppet's [arithmetic operators.][arithmetic]
...
 
 Integers are numbers without decimal points.
 
-If you divide two integers, the result _will not_ be a float; instead, Puppet will truncate the remainder. (That is, `2/3 == 0`.)
+If you divide two integers, the result _is not_ a floating point number. Instead, Puppet truncates the remainder. For example, `2/3 == 0`.
 
 ### Floating point numbers
 
-Floating point numbers ("floats") are numbers that include a fractional value after a decimal point (even if that fractional value is zero, like `2.0`).
+Floating point numbers ("floats") are numbers that include a fractional value after a decimal point, even if that fractional value is zero (such as `2.0`).
```
thallgren: A number with an exponent (scientific notation) such as 2e32 is also a float even if it doesn't have decimal point. E.g. the following are all legal floats
```
1.2
12e-1
12.32e1
-12.32e+2
```

source/puppet/4.10/lang_data_number.md
```diff
@@ -57,49 +56,63 @@ $product = 8 * 3e5  # evaluates to 240000.0
...
+-   Create a new `Integer` or `Float` object using the string as a parameter, such as `$mynum = Integer($mystring)`. This invokes the type's `new()` function, which can convert objects of other types --- including Strings --- into [Integers](./function.html#conversion-to-integer) and [Floats](./function.html#conversion-to-float).
+
+    The default radix of this conversion is base 10, but you can specify others.
+-   Use [the `scanf` function](./function.html#scanf) to manually extract numbers from strings. This function can also account for surrounding non-numerical text.
+
+> **Deprecation note:** Puppet 4.10.2 deprecated numeric coercion, which silently and automatically converted strings into numbers by adding a number to a string (for example, `$mynum = '10' + 0` would result in `$mynum` being assigned the number 10.) This now logs a Warning, or generates an Error if `--strict` is enabled.
+>
+> To ensure that your code works with future versions of Puppet, create new instances of numeric objects to convert String values to numeric types.
```
hlindberg: Or - (obviously), avoid having numbers in string form in the first place... then there is no need to convert anything.

source/puppet/4.10/lang_data_number.md
```diff
@@ -57,49 +56,63 @@ $product = 8 * 3e5  # evaluates to 240000.0
...
+
+### Examples
+
+Converting a string to an integer and float:
 
+    $mystring = '10';                  # $mystring == '10'
+    $myinteger = Integer($mystring);   # $myinteger == 10
+    $myfloat = Integer($mystring);     # $myfloat == 10.0
```
hlindberg: nope, that creates an Integer, not a Float, did you mean `Float($mystring)` ?

source/puppet/4.10/lang_data_number.md
```diff
@@ -57,49 +56,63 @@ $product = 8 * 3e5  # evaluates to 240000.0
...
+    $myhexint = Integer($mystring, 16) # $myhexint == 16; '10' in base 16
 
 ## The `Integer` data type
 
 The [data type][] of integers is `Integer`.
 
 By default, `Integer` matches whole numbers of any size (within the limits of available memory).
```
hlindberg: That is a strange description of what an Integer matches - has nothing to do with the limits of available memory. An integer is specified to be a 64 bit value and that sets the limits of which whole numbers it can match.

source/puppet/4.10/lang_data_number.md
```diff
@@ -109,28 +122,27 @@ The full signature for `Integer` is:
...
----------| -----------------|-----------|---------------|------------
-1 | Min value | `Integer` | negative infinity | The minimum value for the integer. This parameter accepts the special value `default`, which will use its default value.
-2 | Max value | `Integer` | infinity | The maximum value for the integer. This parameter accepts the special value `default`, which will use its default value.
+Position | Parameter        | Data type | Default value     | Description
+---------| -----------------|-----------|-------------------|-----------------------------
+1        | Minimum value    | `Integer` | negative infinity | The minimum value for the integer. This parameter accepts the special value `default`, which uses its default value.
+2        | Maximum value    | `Integer` | infinity          | The maximum value for the integer. This parameter accepts the special value `default`, which uses its default value.
 
 Practically speaking, the integer size limit is the range of a 64-bit signed integer (−9,223,372,036,854,775,808 to 9,223,372,036,854,775,807), which is the maximum size that can roundtrip safely between the components in the Puppet ecosystem.
```
hlindberg: The term "size" is awkward here - since that is a positive number - the number of distinct values in the range of values a 64 bit integer can describe. How about saying "the range of integer values" instead?

source/puppet/4.10/lang_data_number.md
```diff
@@ -109,28 +122,27 @@ The full signature for `Integer` is:
...
 All of these parameters are optional. They must be listed in order; if you need to specify a later parameter, you must specify values for any prior ones.
 
-Position | Parameter        | Data Type | Default Value | Description
----------| -----------------|-----------|---------------|------------
-1 | Min value | `Integer` | negative infinity | The minimum value for the integer. This parameter accepts the special value `default`, which will use its default value.
-2 | Max value | `Integer` | infinity | The maximum value for the integer. This parameter accepts the special value `default`, which will use its default value.
+Position | Parameter        | Data type | Default value     | Description
+---------| -----------------|-----------|-------------------|-----------------------------
+1        | Minimum value    | `Integer` | negative infinity | The minimum value for the integer. This parameter accepts the special value `default`, which uses its default value.
+2        | Maximum value    | `Integer` | infinity          | The maximum value for the integer. This parameter accepts the special value `default`, which uses its default value.
```
hlindberg: it isn't really infinity and negative infinity, it is MAXINT, and MININT the respective max/min a 64 bit integer can describe. (The values shown below).

source/puppet/4.10/lang_data_number.md
```diff
@@ -140,26 +152,23 @@ The full signature for `Float` is:
...
 
-Position | Parameter        | Data Type | Default Value | Description
----------| -----------------|-----------|---------------|------------
-1 | Min value | `Float` | negative infinity | The minimum value for the float. This parameter accepts the special value `default`, which will use its default value.
-2 | Max value | `Float` | infinity | The maximum value for the float. This parameter accepts the special value `default`, which will use its default value.
-
+Position | Parameter        | Data type | Default value     | Description
+---------| -----------------|-----------|-------------------|-----------------------------
+1        | Minimum value    | `Float`   | negative infinity | The minimum value for the float. This parameter accepts the special value `default`, which uses its default value.
+2        | Maximum value    | `Float`   | infinity          | The maximum value for the float. This parameter accepts the special value `default`, which uses its default value.
```
hlindberg: Note - here it is more correct to use "infinity" and "negative infinity", but there is a MIN and MAX for float values as well as specified in the language specification - it is an IEEE standard for 64 bit floating point values. The details are perhaps not very relevant to a puppet programmer since puppet isn't used for calculations where this matters, but could be good to know that you may run against these limits if you multiply some very large floating point numbers. Perhaps just refer to the IEEE standard?

hlindberg:
My comments are for 4.10 version but applies to all of the versions

gguillotte:
These are great comments @hlindberg @thallgren, though most are on changes to formatting and not content, which means those things have been wrong for a very long time. I'll address them in follow-up commits, but it will take some time.

PR #854, opened 2018-02-26
Add `puppet-agent` component list up to v5.4.0

sigv:
#810 has happened again.

The latest version on the list is [`5.3.4`](https://github.com/puppetlabs/puppet-docs/blob/342c66fd6e7e6cb4c741c38855f79dd09da98e00/source/puppet-agent/_agent5.x.html#L9) even though there are tags for [`5.3.5`](https://github.com/puppetlabs/puppet-agent/tree/5.3.5) and [`5.4.0`](https://github.com/puppetlabs/puppet-agent/tree/5.4.0) already.


source/puppet-agent/_agent5.x.html
```diff
@@ -6,6 +6,8 @@
   </thead>
 
   <tbody>
+    <tr><td><a href="/puppet/5.4/release_notes_agent.html#puppet-agent-540">5.4.0</a></td> <td><a href="/puppet/5.4/release_notes.html#puppet-540">5.4.0</a></td> <td><a href="/facter/3.10/release_notes.html#facter-3100">3.10.0</a></td> <td><a href="/hiera/3.4/release_notes.html#hiera-342">3.4.2</a></td> <td><a href="/mcollective/current/releasenotes.html">2.11.4</a></td> <td>2.4.3</td> <td>1.0.2n</td></tr>
```
sigv: Either [/puppet/5.4/release_notes_agent.md](https://github.com/puppetlabs/puppet-docs/blob/e5fbccccbefac8909063355b6867025b61fb89e3/source/puppet/5.4/release_notes_agent.md#puppet-agent-540) is wrong or I am reading this wrong, but [puppet-agent@5.4.0 /configs/components/marionette-collective.json](https://github.com/puppetlabs/puppet-agent/blob/5.4.0/configs/components/marionette-collective.json) is set to track `2.11.4` instead of the latest `2.12.0` as the documentation linked suggests, therefore I have set it that way.

sigv: Added change for Release Notes regarding this.

sigv:
@gguillotte, direct ping.

gguillotte:
Been out sick, thanks for the heads up @sigv.

gguillotte:
CC @branan for data points toward docs automation

source/puppet-agent/_agent5.x.html
gguillotte: CC @khussey, looks like the source of truth for MCo was updated after the release notes were compiled.

sigv: Raises the question what the source of truth is considered if not the tag.. or are you saying the tag was rewritten?

gguillotte: @sigv We have an internal pre-release wiki page where we plan which components will be in each release. It looks like there was a late change to MCo that I missed when compiling the release notes prior to the release.

PR #856, opened 2018-03-05
(DOCUMENT-833) Clarify fact types and flat/structured values

gguillotte:
From the ticket:

> In a discussion with a Puppet User, it became clear that our documentation about facts is a little unclear. Specifically the concept I had to explain was that all types of fact (standard, executable or external) can return either simple or structured output.

> Here is the URL under discussion: https://puppet.com/docs/facter/3.10/custom_facts.html


gguillotte:
@MikaelSmith Am I on the right track here? I'm not entirely confident about some of the terms I'm using here based on existing text, particularly "flat" facts to describe non-structured facts, and the segmentation of core, custom (Ruby), and external (structured data and executable) facts.

PR #857, opened 2018-03-12
Fix OS family fact value

pmuller:
The fact value exposed by facter on RHEL-like systems is "RedHat", not "redhat".


gguillotte:
Thanks, @pmuller!

PR #869, opened 2018-05-15
expanding reference.md info, editing style

jbondpdx:


source/puppet/5.5/modules_documentation.markdown
```diff
@@ -194,116 +194,103 @@ To transition your module to the `REFERENCE.md` model, add Strings comments to y
...
   * Valid values, if the data type doesn't make it obvious.
   * Default value, if any.
 
+```markdown
+# Reference
+
+## Classes
+
+### Public Classes
+*[`cat`](#::cat)
```
eputnam: unless :: is in the heading, it doesn't need to be included in the link, this can just be (#cat). i'm just afraid this could be misinterpreted as a need to incorporate namespacing into your documentation :)

source/puppet/5.5/modules_documentation.markdown
```diff
@@ -194,116 +194,103 @@ To transition your module to the `REFERENCE.md` model, add Strings comments to y
...
 To write a clear, concise, and comprehensible documentation, we recommend following a few principles, best practices, and style guidelines.
 
-{:.section}
-### General principles of READMEs
-
-1. Write for both web and terminal viewing. We recommend [Markdown][commonmark]. Above all else, your module must be easily readable and scannable.
-2. Limit the number of external links. Linking to anything on the web limits the usability of the module, particularly if a range of users might use it in various environments, such as in terminal. Also, links look gross in plain Markdown and make your README less readable.
-3. Scannability is key. READMEs are formulaic and repetitious for a reason. Repetition means that no matter the module, users know where to get the information they're looking for.
+1. Write for both web and terminal viewing using [Markdown][commonmark]. Above all else, your module must be easily readable and scannable.
+1. Limit the number of external links. Linking to anything on the web limits the usability of the module, particularly when users read it in terminal.
```
eputnam: Something else I found along these same lines was that local links, or links to other docs in the module, will not work in the Forge display. The biggest example being linking from the Reference section in the README to the REFERENCE.md file. But also if you try to include links from REFERENCE.md to manifests or other files, they won't work on the Forge view. Dunno if we want to mention this.

source/puppet/5.5/modules_documentation.markdown
```diff
@@ -194,116 +194,103 @@ To transition your module to the `REFERENCE.md` model, add Strings comments to y
...
-
-### Classes
-
-#### Public classes
-
-*[`pet::cat`](#petcat): Installs and configures a cat in your environment.
+If you already have a README with an accurate reference section, you can remove that section from your README and paste it into a `REFERENCE.md` file. The formatting isn't vastly different.
 
-#### Private classes
+> **Important**: If you create a `REFERENCE.md` manually, and then generate Markdown with Puppet Strings, it will overwrite your `REFERENCE.md` unless you give the files different name. The module detail page on the Forge displays only the `README.md` and `REFERENCE.md` files.
```
eputnam: "...it will overwrite your `REFERENCE.md` unless you give the files different name*s*."?

source/puppet/5.5/modules_documentation.markdown
jbondpdx: Thanks! It's old and leftover; I'll fix it.

source/puppet/5.5/modules_documentation.markdown
jbondpdx: Should this be a Forge ticket to fix? Or is it just "won't work"?

source/puppet/5.5/modules_documentation.markdown
jbondpdx: Oops! I kept going back and forth, is what happened.

source/puppet/5.5/modules_documentation.markdown
jbondpdx: Filing FORGE-429 against this issue.

PR #878, opened 2018-07-02
(WEB-2872) Restore some archived content

gguillotte:
To migrate some content to from docs.puppet.com to puppet.com/docs, we need to restore it from the `archive` branch and to `_config.yml`.


gguillotte:
@jtappa I've built content locally without errors but would appreciate a sanity check on the _config.yml changes. All the content restored into the repo was copied from the `archive` branch, and the only content changes were to remove the `yard` directories from Puppet 3.8 and 4.5 and replace them with the YARD generation instructions.

PR #881, opened 2018-08-01
(DOCUMENT-719) Draft changes to major Server version upgrade docs

gguillotte:
There are concerns about the order of steps in the major version upgrade docs from Puppet 3 to Puppet 5, particularly around PuppetDB upgrades now being necessary rather than optional. See [DOCUMENT-719](https://tickets.puppetlabs.com/browse/DOCUMENT-719) for details.

### DIFF
--- a/source/puppet/5.5/upgrade_major_server.markdown
+++ b/source/puppet/5.5/upgrade_major_server.markdown
@@ -52,16 +52,6 @@ To upgrade Puppet Server, you'll need to add the Puppet Platform repository to e
 
 Even after you've installed the package, **don't start the `puppetserver` service yet**! You should do a few other things first.
 
-## Upgrade PuppetDB
-
-In the pre-upgrade steps, you upgraded to PuppetDB 2.3.x, including the [`puppetdb-terminus`][puppetdb-terminus] package on your Puppet Servers. This means the upgraded Puppet Server can already communicate with your existing PuppetDB server.
-
-Now that you've upgraded Puppet Server, [upgrade PuppetDB][] to the latest version 3.x. Note that this also retires older API versions, which can break older integrations. See the [PuppetDB 3.0 release notes][] for details.
-
-Use the [`puppetlabs/puppetdb`][puppetdb_module] module to manage your PuppetDB version. Also, note that the terminus package's name is now [`puppetdb-termini`][puppetdb-termini] instead of `puppetdb-terminus`.
-
-After successfully upgrading PuppetDB to version 3, upgrade it to the latest v4 edition of PuppetDB and `puppetdb-termini`.
-
 ### Get familiar with the new binary locations
 
 In Puppet 4, we [moved][] all of Puppet's binaries on \*nix systems. They are now at `/opt/puppetlabs/bin`, which isn't in your system's `PATH` by default. You should do one of the following:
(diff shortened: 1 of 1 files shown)


jbondpdx:
@clairecadman since you are the current Server writer, could you please review this PR, and make sure any necessary updates are also made to the Puppet 6 docs? Thank you!

source/puppet/5.5/upgrade_major_server.markdown
```diff
@@ -44,14 +44,24 @@ If you have multiple Puppet masters, upgrade or replace the certificate authorit
...
 
+## Upgrade PuppetDB
+
+In the pre-upgrade steps, you upgraded to PuppetDB 2.3.x, including the [`puppetdb-terminus`][puppetdb-terminus] package on your Puppet Servers. This means the upgraded Puppet Server can already communicate with your existing PuppetDB server.
+
+Now that you've upgraded Puppet Server, [upgrade PuppetDB][] to the latest version 3.x. Note that this also retires older API versions, which can break older integrations. See the [PuppetDB 3.0 release notes][] for details.
+
+Use the [`puppetlabs/puppetdb`][puppetdb_module] module to manage your PuppetDB version. Also, note that the terminus package's name is now [`puppetdb-termini`][puppetdb-termini] instead of `puppetdb-terminus`.
+
+After successfully upgrading PuppetDB to version 3, upgrade it to the latest v4 edition of PuppetDB and `puppetdb-termini`.
```
justinstoller: Does this need to happen after Puppet Server has been started? (I assume you'd be using `puppet apply` if not?). If so we might want to move the section down a bit? Or split it between what is required before starting Puppet Server and what is done after starting it.

source/puppet/5.5/upgrade_major_server.markdown
```diff
@@ -44,14 +44,24 @@ If you have multiple Puppet masters, upgrade or replace the certificate authorit
 
 Repeat the following steps for each Puppet Server until you're running a pure Puppet 5/Puppet Server 5 infrastructure.
 
-### Install the latest Puppet Server
+### Install the latest Puppet Server 2.x
```
justinstoller: Can we mention that this is only required if on Puppet Server < 2.latest either in the title or below? (I might be overly concerned that someone will first downgrade to 2.8 from 5.x)

source/puppet/5.5/upgrade_major_server.markdown
clairecadman: @gguillotte Are you happy if I move this below Start Puppet Server? Or do you think it should be split out?

source/puppet/5.5/upgrade_major_server.markdown
clairecadman: @justinstoller Do you mean on Puppet Server 2.0 and earlier or later?

clairecadman:
@rbrw will review the comments on line 63 sometime this week, and then I'll go ahead and merge.  Thanks all!

source/puppet/5.5/upgrade_major_server.markdown
```diff
@@ -200,17 +210,15 @@ Log into any Puppet agent and test its connection to the upgraded Puppet Server:
...
-### Go live!
-
-At this point, Puppet Server is ready to serve nodes in production. If you pulled the server back to stage the upgrade, you can now push the node back into use.
+### Upgrade to Puppet Platform 5
 
-##  Optional: Upgrade PuppetDB
+Compared to the upgrade from Puppet 3 to 4, there are fewer changes when upgrading between Puppet 4 and 5. See the [Puppet 5.0 release notes](../5.0/release_notes.html), [Puppet Server 5.0 release notes](/docs/puppetserver/5.0/release_notes.html), and [PuppetDB 5.0 release notes](/docs/puppetdb/5.0/release_notes.html) for details, then review the latest version of each product's release notes.
 
-In the pre-upgrade steps, you upgraded to PuppetDB 2.3.x, including the [`puppetdb-terminus`][puppetdb-terminus] package on your Puppet Servers. This means the upgraded Puppet Server can already communicate with your existing PuppetDB server.
+Finally, upgrade PuppetDB the latest version of PuppetDB 5, then upgrade Puppet Server and `puppetdb-termini` together on each Puppet master.
```
rbrw: "upgrade PuppetDB to"?

source/puppet/5.5/upgrade_major_server.markdown
rbrw: I think the general rule has been to upgrade PuppetDB before Puppet Server and the terminus, so that it's ready to handle data/requests from both older and newer server/terminus versions.

I don't think that Puppet Server has to be stopped as far as PuppetDB itself is concerned, you just want to avoid a "too new" Puppet Server sending requests to a PuppetDB that can't handle them.

gguillotte: Is it a problem for PuppetDB to be newer than Server? I thought avoiding that "too new" Server problem was part of the point of upgrading PDB + termini together from 2.3.latest to 3.latest, then 4.latest, before starting Server 2.latest.

PR #882, opened 2018-08-09
(DOC-3625) Update docs that link to type reference

gguillotte:
When Puppet docs link to reference, prefer links to the
individual types over the single-page reference.

This PR also updates Markdown formatting and removes some outdated
content to help prepare it for migration.


gguillotte:
@binford2k If there are content issues outside of the scope of links to the type reference, please file additional DOC tickets so the team can triage them.

jbondpdx:
Modules fundamentals has already been migrated, but the rest of these are still in play. Thank you!

PR #883, opened 2018-08-10
(DOC-3645) Update class name regex.

gguillotte:
Updates the class name regex and cleans up some Markdown formatting.

### DIFF
--- a/source/puppet/5.5/lang_reserved.markdown
+++ b/source/puppet/5.5/lang_reserved.markdown
@@ -21,7 +21,6 @@ title: "Language: Reserved words and acceptable names"
 
 ## Reserved words
 
-
 Several words in the Puppet language are **reserved**. This means they:
 
 * Cannot be used as bare word strings --- you must quote these words if you wish to use them as strings.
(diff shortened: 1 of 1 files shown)


source/puppet/5.5/lang_reserved.markdown
```diff
@@ -138,15 +132,15 @@ The names of classes and defined resource types can consist of one or more [name
...
+    \A[a-z][a-zA-Z0-9_]*\Z
 
 The one exception is the top namespace, whose name is the empty string.
 
 Multiple namespace segments can be joined together in a class or defined type name with the `::` (double colon) [namespace][] separator.
 
 Class names with multiple namespaces should match the following regular expression:
 
-    \A([a-z][a-z0-9_]*)?(::[a-z][a-z0-9_]*)*\Z
+    \A([a-z][a-zA-Z0-9_]*)?(::[a-z][a-zA-Z0-9_]*)*\Z
```
binford2k: The actual regex used in the codebase is `\A((::)?[a-z]\w*)(::[a-z]\w*)*\z`. As far as I can tell, this is equivalent.

https://github.com/puppetlabs/puppet/blob/952414c36b45df9a0f46f4fe816bac4d2b18a89e/lib/puppet/pops/patterns.rb#L26

gguillotte: Pushed a revision, and also emphasized that these are Ruby regex.

PR #884, opened 2018-08-14
(DOC-3716) Document Timestamp and Timespan types

gguillotte:
Puppet 4 added Timestamp and Timespan types, but were not documented outside of the `new()` function and [Puppet language specification](https://github.com/puppetlabs/puppet-specifications/blob/master/language/expressions.md#timespan-type--). Add documentation for these types to the abstract data type docs.

### DIFF
--- a/source/puppet/5.5/lang_data_abstract.md
+++ b/source/puppet/5.5/lang_data_abstract.md
@@ -117,6 +117,40 @@ Note that you can use capture groups in the regular expressions, but they won't 
 * `Pattern[/\A[a-z].*/]` --- matches any string that begins with a lowercase letter.
 * `Pattern[/\A[a-z].*/, /\Anone\Z/]` --- matches the above **or** the exact string `"none"`.
 
+### `Timestamp`
+
+The `Timestamp` data type matches special [strings][], [floats][], or [integers][] representing a time and date, or `default`, representing an infinite point in the past or future.
+
+It takes up to two parameters, and requires at least one. A `Timestamp` with one parameter represents a single point in time, while two parameters represent a range of time, with the first parameter being the `from` value and the second being the `to` value.
+
+#### Parameters
+
+The full signature for `Timestamp` is:
+
+    Timestamp[ <TIMESTAMP VALUE>, (<RANGE LIMIT>) ]
+
+Position | Parameter | Data Type | Default Value | Description
+---------|-----------|-----------|---------------|------------
+1 | Timestamp Value | String, Float, Integer, or `default` | `default` (-Infinity in a range) | Point in time if passed alone, or `from` value in a range if passed with a second parameter
+2 | Range Limit | String, Float, Integer, or `default` | `default` (+Infinity) | The `to` value in a range
+
+`Timestamp` values are interpreted depending on their format.
+
+* A String in the format of `YYYY-MM-DD` represents the time 00:00:00.0 UTC on the specified year (`YYYY`), month (`MM`), and day (`DD`)
+* A String in the format of `YYYY-MM-DD`T`HH:MM:SS.SSS` represents a specific hour (`HH`), minute (`MM`), and decimal second (`SS.SSS`) of UTC on the specified date
+* An Integer or Float represents the seconds passed since the Unix epoch
+* `default` is used in ranges and represents infinity as the maximum floating point value, either negative (as the `from` value of a range) or positive (as the `to` value)
+
+A `Timestamp` range with either end defined as `default` (infinity) is an _open range_, while a range with both ends defined as an exact timestamp is a _closed range_. Range values can be negative.
+
+The `Timestamp` type is not enumerable.
+
+#### Examples:
+
+* `Timestamp['2000-01-01T00:00:00.000', default]` --- an open range of time from the start of the 21st century to an infinte point in the future
+* `Timestamp['2012-10-10']` --- the exact Timestamp 2012-10-10T00:00:00.0 UTC
+* `Timestamp[default, 1433116800]` --- an open range of time from an infinite point in the past to 2015-06-01T00:00:00 UTC, here expressed as seconds since the Unix epoch
+* `Timestamp['2010-01-01', '2015-12-31T23:59:59.999999999']` --- a closed range of time between the start of 2010 and the end of 2015
 
 ### `Enum`
 
@@ -281,7 +315,6 @@ It takes no parameters.
 ### `Any`
 
 The `Any` data type matches _any_ value of _any_ data type.
-
 
 ## Unusual types
 
--- a/source/puppet/5.5/lang_data_abstract.md
+++ b/source/puppet/5.5/lang_data_abstract.md
@@ -119,7 +119,7 @@ Note that you can use capture groups in the regular expressions, but they won't 
 
 ### `Timespan`
 
-The `Timespan` data type matches a Timespan; [strings][], [floats][], or [integers][] that represent a time and date; or `default`, which represents an infinite point in the past or future.
+The `Timespan` data type matches a Timespan; [strings][], or [numbers][] that represent a time and date; or `default`, which represents an infinite point in the past or future.
 
 It takes up to two parameters, and defaults to an infinite range to the past and future.
 
@@ -134,10 +134,14 @@ Position | Parameter | Data Type | Default Value | Description
 1 | Timespan Start or Length | String, Float, Integer, or `default` | `default` (-Infinity in a span) | Length of the timespan if passed alone, or the `from` value in a range if passed with a second parameter
 2 | Timespan End | String, Float, Integer, or `default` | `default` (+Infinity), or none if only one value is passed | The `to` value in a range
 
-`Timeframe` values are interpreted depending on their format.
+`Timespan` values are interpreted depending on their format.
 
 * A String in the format of `D-HH:MM:SS` represents a span of days (`D`), hours (`HH`), minutes (`MM`), and seconds (`SS`)
 * An Integer or Float represents a number of seconds
+
+A `Timespan` range with either end defined as `default` (infinity) is an _open range_, while any other span is a _closed range_. The span's range is inclusive.
+
+The `Timespan` type is not enumerable.
 
 #### Examples:
 
@@ -147,7 +151,7 @@ Position | Parameter | Data Type | Default Value | Description
 
 ### `Timestamp`
 
-The `Timestamp` data type matches a Timestamp; [strings][], [floats][], or [integers][] that represent a time and date; or `default`, which represents an infinite point in the past or future.
+The `Timestamp` data type matches a Timestamp; [strings][] or [numbers][] that represent a time and date; or `default`, which represents an infinite point in the past or future.
 
 It takes up to two parameters, and defaults to an infinite range to the past and future. A `Timestamp` with one parameter represents a single point in time, while two parameters represent a range of time, with the first parameter being the `from` value and the second being the `to` value.
 
@@ -160,7 +164,7 @@ The full signature for `Timestamp` is:
 Position | Parameter | Data Type | Default Value | Description
 ---------|-----------|-----------|---------------|------------
 1 | Timestamp Value | String, Float, Integer, or `default` | `default` (-Infinity in a range) | Point in time if passed alone, or `from` value in a range if passed with a second parameter
-2 | Range Limit | String, Float, Integer, or `default` | `default` (+Infinity) | The `to` value in a range
+2 | Range Limit | String, Float, Integer, or `default` | `default` (+Infinity), or none if only one value is passed | The `to` value in a range
 
 `Timestamp` values are interpreted depending on their format.
 
@@ -169,7 +173,7 @@ Position | Parameter | Data Type | Default Value | Description
 * An Integer or Float represents the seconds passed since the Unix epoch
 * `default` is used in ranges and represents infinity as the maximum floating point value, either negative (as the `from` value of a range) or positive (as the `to` value)
 
-A `Timestamp` range with either end defined as `default` (infinity) is an _open range_, while a range with both ends defined as an exact timestamp is a _closed range_. Range values can be negative.
+A `Timestamp` range with either end defined as `default` (infinity) is an _open range_, while a range with both ends defined as an exact timestamp is a _closed range_. Range values can be negative, and the range is inclusive.
 
 The `Timestamp` type is not enumerable.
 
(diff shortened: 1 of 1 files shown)


jbondpdx:
@thallgren , we are blocked on this review. Could you have a look at this and see if it's OK to merge? Thank you!

jbondpdx:
@hlindberg , can you help out with a review on this? I'd like to get this into the docs.

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* A String in the format of `D-HH:MM:SS` represents a span of days (`D`), hours (`HH`), minutes (`MM`), and seconds (`SS`)
+* An Integer or Float represents a number of seconds
+
+A `Timespan` range with either end defined as `default` (infinity) is an _open range_, while any other span is a _closed range_. The span's range is inclusive.
+
+The `Timespan` type is not enumerable.
+
+#### Examples:
+
+* `Timespan[2]` --- a timespan of 2 seconds
```
hlindberg:
```suggestion
* `Timespan[2]` --- matches a Timespan value of 2 seconds
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* `Timespan[77.3]` --- a timespan of 1 minute, 17 seconds, and 300 milliseconds (77.3 seconds)
```
hlindberg:
```suggestion
* `Timespan[77.3]` --- matches a Timespan value of 1 minute, 17 seconds, and 300 milliseconds (77.3 seconds)
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* `Timespan['1-00:00:00', '2-00:00:00']` --- a timespan between 1 and 2 days
```
hlindberg:
```suggestion
* `Timespan['1-00:00:00', '2-00:00:00']` --- matches  Timespan values between 1 and 2 days
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
 * `Pattern[/\A[a-z].*/]` --- matches any string that begins with a lowercase letter.
 * `Pattern[/\A[a-z].*/, /\Anone\Z/]` --- matches the above **or** the exact string `"none"`.
 
+### `Timespan`
+
+The `Timespan` data type matches a Timespan; [strings][], or [numbers][] that represent a time and date; or `default`, which represents an infinite point in the past or future.
```
hlindberg:
```suggestion
The `Timespan` data type matches Timespan values within a given range; specified with [strings][] or [numbers][] that represent duration; or `default`, which represents a positive or infinite duration.
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+
+#### Examples:
+
+* `Timespan[2]` --- a timespan of 2 seconds
+* `Timespan[77.3]` --- a timespan of 1 minute, 17 seconds, and 300 milliseconds (77.3 seconds)
+* `Timespan['1-00:00:00', '2-00:00:00']` --- a timespan between 1 and 2 days
+
+### `Timestamp`
+
+The `Timestamp` data type matches a Timestamp; [strings][] or [numbers][] that represent a time and date; or `default`, which represents an infinite point in the past or future.
```
hlindberg: This has the same issues as for `Timespan` - a `Timestamp` only matches `Timestamp` values.

hlindberg:
```suggestion
The `Timestamp` data type matches Timestamp values within a given range; specified with [strings][] or [numbers][] that represents the points in time limiting the range; or `default`, which represents an infinite point in the past or future
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+
+It takes up to two parameters, and defaults to an infinite range to the past and future. A `Timestamp` with one parameter represents a single point in time, while two parameters represent a range of time, with the first parameter being the `from` value and the second being the `to` value.
```
hlindberg:
```suggestion
It takes up to two parameters, and defaults to an infinite range to the past and future. A `Timestamp` with one parameter represents a single point in time, while two parameters represent a range of timestamps, with the first parameter being the `from` value and the second being the `to` value.
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* An Integer or Float represents the seconds passed since the Unix epoch
+* `default` is used in ranges and represents infinity as the maximum floating point value, either negative (as the `from` value of a range) or positive (as the `to` value)
+
+A `Timestamp` range with either end defined as `default` (infinity) is an _open range_, while a range with both ends defined as an exact timestamp is a _closed range_. Range values can be negative, and the range is inclusive.
+
+The `Timestamp` type is not enumerable.
+
+#### Examples:
+
+* `Timestamp['2000-01-01T00:00:00.000', default]` --- an open range of time from the start of the 21st century to an infinte point in the future
```
hlindberg:
```suggestion
* `Timestamp['2000-01-01T00:00:00.000', default]` --- an open range of timestamps from the start of the 21st century to an infinte point in the future
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* `Timestamp['2012-10-10']` --- the exact Timestamp 2012-10-10T00:00:00.0 UTC
```
hlindberg:
```suggestion
* `Timestamp['2012-10-10']` --- matches the exact Timestamp 2012-10-10T00:00:00.0 UTC
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* `Timestamp[default, 1433116800]` --- an open range of time from an infinite point in the past to 2015-06-01T00:00:00 UTC, here expressed as seconds since the Unix epoch
```
hlindberg:
```suggestion
* `Timestamp[default, 1433116800]` --- an open range of timestamps from an infinite point in the past to 2015-06-01T00:00:00 UTC, here expressed as seconds since the Unix epoch
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* `Timestamp['2010-01-01', '2015-12-31T23:59:59.999999999']` --- a closed range of time between the start of 2010 and the end of 2015
```
hlindberg:
```suggestion
* `Timestamp['2010-01-01', '2015-12-31T23:59:59.999999999']` --- a closed range of timestamps between the start of 2010 and the end of 2015
```

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+
+`Timespan` values are interpreted depending on their format.
+
+* A String in the format of `D-HH:MM:SS` represents a span of days (`D`), hours (`HH`), minutes (`MM`), and seconds (`SS`)
+* An Integer or Float represents a number of seconds
+
+A `Timespan` range with either end defined as `default` (infinity) is an _open range_, while any other span is a _closed range_. The span's range is inclusive.
+
+The `Timespan` type is not enumerable.
+
```
hlindberg: May want to link to the documentation for Timspan.new https://puppet.com/docs/puppet/6.0/function.html#conversion-to-timespan

source/puppet/5.5/lang_data_abstract.md
```diff
@@ -117,6 +117,72 @@ Note that you can use capture groups in the regular expressions, but they won't
...
+* A String in the format of `YYYY-MM-DD` represents the time 00:00:00.0 UTC on the specified year (`YYYY`), month (`MM`), and day (`DD`)
+* A String in the format of `YYYY-MM-DD`T`HH:MM:SS.SSS` represents a specific hour (`HH`), minute (`MM`), and decimal second (`SS.SSS`) of UTC on the specified date
+* An Integer or Float represents the seconds passed since the Unix epoch
+* `default` is used in ranges and represents infinity as the maximum floating point value, either negative (as the `from` value of a range) or positive (as the `to` value)
+
+A `Timestamp` range with either end defined as `default` (infinity) is an _open range_, while a range with both ends defined as an exact timestamp is a _closed range_. Range values can be negative, and the range is inclusive.
+
+The `Timestamp` type is not enumerable.
+
```
hlindberg: May want to link to the documentation for Timspan.new https://puppet.com/docs/puppet/6.0/function.html#conversion-to-timestamp

hlindberg:
@gguillotte ok - `Timestamp` and `Timespan` are indeed scalars (`Timestamp =~  Type[Scalar]` is `true`) and the same for `Timespan`. If that means a page each, or one with "other scalars", or one with "time related scalars" is totally up to you. My point here was that they are not "abstract data types" like for example `Enum` (the term "abstract" means that there are never any instances of that data type).

The specification repository has more text about the `Timestamp` and `Timespan` data types, it shows how they are created etc. Is that enough as predocs?

gguillotte:
@hlindberg Knowing how to create data in these types is useful, but the spec doesn't explain when, where, why, or how these types would be most useful, and I wouldn't expect it to—it's a language specification, not a user-focused document. The `new()` docs explain how to convert values to `Timespan` and `Timestamp` but also doesn't explain why converting to these types is useful. The description of PUP-5871 indicates that they're useful types to have, but not why or how.

I don't have the context to answer these questions, so I think user-focused pre-docs would help the writer taking on this responsibility going forward. This PR's goal is to fill a pretty gaping hole of minimum documentation about types that were introduced two major versions ago, but I with the context I have from reading the spec and ticket, I can't explain to someone in detail what problems they were designed and implemented to solve.

Additionally, there are other docs that user-focused pre-docs could help the writer update with these types. For example:

- Are `Timespan` and `Timestamp` useful in EPP/ERB templates, and if so, are there particularly valuable uses in that context? The spec doesn't cover that, and user-focused pre-docs would help update `lang_template_epp` and `_erb`.

- Do any core facts already use `Timespan` and `Timestamp`? The spec doesn't describe any interactions or core facts using them, but if they do, user-focused pre-docs would help update `lang_facts_and_builtin_vars`.

- How are `Timespan` and `Timestamp` converted in custom `Puppet::Function` Ruby functions? I'm not sure how to interpret the [Pcore generic data transformations](https://github.com/puppetlabs/puppet-specifications/blob/6655ac6bf996268d625c4d985c3d707ec0f7ff3b/language/data-types/pcore-generic-data.md) in a user context—it shows `Timestamp` is converted to a JSON object but `functions_ruby_overview` doesn't describe that behavior, and `Timespan` isn't described at all. User-focused pre-docs about using time-related types in custom Ruby functions would help.

Those questions aren't going to be answered in this PR, but pre-docs that could help describe these types in this PR could also help answer those questions for further user-focused improvements.

hlindberg:
Well..

> Are Timespan and Timestamp useful in EPP/ERB templates, and if so, are there particularly valuable uses in that context?

Obviously if you have data values that are dealing with time/date and/or duration and you need to format them in a particular format for a particular kind of file.
Also if you need to do any calculations like taking user input (in some format) and adding or subtracting time from it to get a new value. That is really difficult if all you have is a string
that describes a date/time.

> The spec doesn't cover that, and user-focused pre-docs would help update lang_template_epp and _erb.
Right, the spec is terse, assumes you would know what to use it for.

>Do any core facts already use Timespan and Timestamp? The spec doesn't describe any interactions or core facts using them, but if they do, user-focused pre-docs would help update ang_facts_and_builtin_vars.

None of the core facts make use of the new data types. Facts do not use rich data yet.

>How are Timespan and Timestamp converted in custom Puppet::Function Ruby functions?

They are not converted. They are first class data types - scalar values. In Puppet 6, rich data is on by default, and if you have types/providers that do support Timespan and Timstamp (probably none, there are none in puppet core since it is a brand new feature that you can use them), you can use them directly in resources, otherwise you would use the various formatting options to get a string out to represent the time/date/duration.

> User-focused pre-docs about using time-related types in custom Ruby functions would help.
Yeah, there is no such docs, except what is inside ruby documentation in the source for those data types.

>Those questions aren't going to be answered in this PR, but pre-docs that could help describe these types in this PR could also help answer those questions for further user-focused >improvements.

 I can certainly help, but maybe get something basic added first (and/or suggestions/tentative-docs or just questions - like the ones here) to make it easier for me to help out.

source/puppet/5.5/lang_data_abstract.md
hlindberg: typo, corrected - should have read "an infinite negative or positive duration" - in practical terms "+ or - the largest possible (i.e. infinite) duration of time", as in "10 minutes or longer" (the upper end is positive infinity), or "you can subtract as much time as you like" (the lower end is negative infinity).

source/puppet/5.5/lang_data_abstract.md
hlindberg: A range of time stamps are bound to points in time; from a specific date/time to another specific date/time. A range of Timestamps is a range of durations. For example - a movie may be 2h30min long (a Timespan), but that does not say on which date and time it is showing (a Timstamp), so a range of timespans may be "movies longer than 60 minutes but shorter than 4h". 

We use `default` in the type system to describe the absolute min/max values of things - we have that already in Integer and Float ranges. It works the same here; the default upper/lower range values can also be explicitly given as literal `default`. Does that make it clearer?

hlindberg: BTW: the term "time" is best avoided, since it is after all relative ;-) and it does not say if you mean a "point in time" (date/time) or a "duration of time" (length). So `Timestamp` is "When", and `Timespan` is "How Long"

source/puppet/5.5/lang_data_abstract.md
hlindberg: I think I answered that in one of the other questions.

gguillotte:
> I can certainly help, but maybe get something basic added first (and/or suggestions/tentative-docs or just questions - like the ones here) to make it easier for me to help out.

Agreed, thanks. I've committed revisions based on the feedback here.

gguillotte:
@hlindberg I would really appreciate it if we could get this merged and allow @clairecadman, the current owner of these docs, to follow up in a PR that she owns if further changes are advisable.

jbondpdx:
+1 to a merge with future followup as needed.

PR #891, opened 2018-09-28
Update lang_data_type.md 

clairecadman:
This commit aims to address the questions in DOCUMENT-537 (https://tickets.puppetlabs.com/browse/DOCUMENT-537) regarding data types.

### DIFF
--- a/source/puppet/6.0/lang_data_type.md
+++ b/source/puppet/6.0/lang_data_type.md
@@ -73,6 +73,10 @@ This is an abstract data type (`Variant`) which takes any number of data types a
 
 Data types are useful in parameter lists, match (`=~`) expressions, case statements, and selector expressions. There are also a few less common uses for them.
 
+### When and how to use data types
+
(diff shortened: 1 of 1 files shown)


source/puppet/6.0/lang_data_type.md
```diff
@@ -73,6 +73,10 @@ This is an abstract data type (`Variant`) which takes any number of data types a
 
 Data types are useful in parameter lists, match (`=~`) expressions, case statements, and selector expressions. There are also a few less common uses for them.
 
+### When and how to use data types
+
+We recommend using data types whenever you can, and aligning them in columns. Class parameters should be typed wherever possible, and the type used should be as specific as possible, for example, using an Enum for input validation, instead of using a String and checking the contents of the string in the code. You have the option to specify String[1] instead of String, because you may want non empty string — this is fine to use but we do not recommend you ensure it. You should also type variables as deep as possible, without affecting readability. If this is not possible, consider creating your own custom data type alias.
```
ahpook: "you may want non empty string" ➡️ "you may want to enforce non-empty strings"
"this is fine to use but we do not recommend you ensure it" ➡️ (can be removed, i think that was a typo in the ticket)
"type variables as deep as possible" ➡️ "specify data types as deeply as possible"

clairecadman:
@ahpook Thanks for your comments! I adjusted it, please merge if you're happy with it.

PR #895, opened 2018-10-09
(DO NOT MERGE)(DOC-3771) Remove `puppet cert` from the list of man pages

gguillotte:
Puppet 6 and onward don't ship a functional `puppet cert` command. However, `/lib/puppet/applications/cert.rb` still exists in the Puppet code base. Force the docs references generator to stop publishing that man page.

Filed as DO NOT MERGE so that this solution can be reviewed by someone with more Ruby experience than me.

---

The references generator is built to categorize a list of commands by grabbing a list of `.rb` files in Puppet's `/lib/puppet/applications/` path. Any commands in the list that don't have a corresponding `.rb` file are ignored, and any `.rb` files not listed are dumped into the "Unknown or new subcommands" section of the [man page index](https://puppet.com/docs/puppet/6.0/man/index.html).

As such, removing `cert` from the list won't remove `puppet cert` from the index page or generating publishing a man page. We have to intercept and remove `cert` from the list of applications built from the list of `.rb` files before man page generation starts.

The problem with doing this is that new versions of Puppet prior to Puppet 6 might still need regenerated references and man pages, and `puppet cert` hasn't been removed from those versions. This PR's change isn't version-sensitive, so this will cause `puppet cert` to be excluded from the generated man pages even if it's for a relevant older Puppet version.

### DIFF
--- a/lib/puppet_references/puppet/man.rb
+++ b/lib/puppet_references/puppet/man.rb
@@ -27,7 +27,6 @@ module PuppetReferences
             core: %w(
           agent
           apply
-          cert
           master
           module
           resource
@@ -122,6 +121,8 @@ EOADDENDUM
         applications = application_files.map {|f| f.basename('.rb').to_s}
         applications.delete('face_base')
         applications.delete('indirection_base')
+        # DOC-3771: Puppet cert still has a Ruby file, but is disabled. Don't build or list a man page for it.
+        applications.delete('cert')
         applications
       end
 


justinstoller:
It looks like `PuppetReferences::Puppet::Man` inherits from `PuppetReferences::Reference` and passes the arguments to that super class. `PuppetReferences::Reference` expects the arg to be a "commit" which it saves to the `@commit` instance variable. That's value used in the yaml header for "built_from_commit". I don't know if that's expected to be an actual sha or the tag? 

If it's the tag then we could match on that to only include these changes for 6+. If it's a sha, things would be a little more challenging because we'd need access to the puppet get repo to see what the tag that sha relates to.

Do you know how this class is called? (ie with a sha or tag?) Do you know how these are actually built? I could answer the sha/tag question if I knew that.

gguillotte:
> Do you know how this class is called? (ie with a sha or tag?)

Either, but usually by tag:

```
rake generate references:puppet VERSION=6.0.2
```

> PuppetReferences::Reference expects the arg to be a "commit" which it saves to the @commit instance variable. ... If it's the tag then we could match on that to only include these changes for 6+.

Perfect! I think that's enough for me to work with. Thanks, @justinstoller

gguillotte:
@justinstoller I think because the generator's leaving those out as expected, it'd be best to leave them in as long as we're still releasing updates to Puppet 5. This seems to be how Nick expected subcommand removal to work when he built this — `cert` is exceptional by persisting in Puppet 6 as a disabled command.

justinstoller:
makes sense

gguillotte:
@justinstoller No such luck, by the time the commit gets to ::Man it's been converted to a sha, and I can't find a way to call back up to the repo from ::Man to .describe() it back into a tag.

joshcooper:
could you `require 'puppet/version'` and then lookup the version and compare?

```ruby
require 'puppet/version'

if Gem::Version.new(Puppet::PUPPETVERSION) < Gem::Version.new("6.0")
```

PR #911, opened 2018-12-13
(PDB-4209) Add section to docs for PuppetDB 6.1.0

Zak-Kent:
Pointing this section at master because 6.1.0 was a foss only release which was
made off of a temporary branch that was on the tip of master.

I think when we do cut a long lived branch we can change the docs to point there for 6.1 and any other foss releases that come out in between so long as they have the same content and we should be fine, but I'm not 100% sure.

### DIFF
--- a/source/_config.yml
+++ b/source/_config.yml
@@ -698,6 +698,15 @@ documents:
       commit: bbc73f6da0a17eac4fbf84758b477ec9dd4cebb2
       subdirectory: documentation
 
+  /puppetdb/6.1:
+    doc: puppetdb
+    version: "6.1"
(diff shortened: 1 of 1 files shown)


Zak-Kent:
Do y'all know if the commit section under external_source can take a commit sha instead of a branch name? If it can maybe we can just use the commit when we tag the temp branch, that way we won't need to keep the branch hanging around after the release.

Zak-Kent:
Closing this pr and will open another one later. The plan currently is to use the tag for the release and then the commit hash for any subsequent docs changes that are needed for the pdb 6.1 docs.

PR #915, opened 2019-02-08
(DOC-3922) Remove mention of the private directory with the ca.pass file

clairecadman:
This commit removes mention of the private directory with the ca.pass file as it won't be read by the CA.


clairecadman:
@Magisus Is this what you meant? (https://tickets.puppetlabs.com/browse/DOC-3922). If so I will also update the latest version of the docs (located in easyDITA). Thanks!

Magisus:
Yeah that looks accurate. We already removed the docs about the `capass` setting, so I guess this is the only spot that's left.

PR #916, opened 2019-02-28
PR'd for questions, don't merge till reviewed. 

jbondpdx:
This PR adds a new Puppet version, a new Puppet Server version, and a new PuppetDB version.

I have a couple of questions, though; see inline comments below.

@katelopresti , I don't know if you can answer my questions, but if not, maybe you can suggest someone who can?

### DIFF
--- a/source/_config.yml
+++ b/source/_config.yml
@@ -318,6 +318,14 @@ documents:
       puppetdb: "4.3"
     hide: true
 
+  /puppet/6.4:
+    doc: puppet
+    version: "6.4"
+    nav: ./_puppet_toc.html
+    my_versions:
+      facter: "3.13"
+      puppetserver: "6.3"
+    hide: true
   /puppet/6.3:
     doc: puppet
     version: "6.3"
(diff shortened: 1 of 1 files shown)


source/_config.yml
```diff
@@ -318,6 +318,14 @@ documents:
       puppetdb: "4.3"
     hide: true
 
+  /puppet/6.4:
+    doc: puppet
+    version: "6.4"
+    nav: ./_puppet_toc.html
+    my_versions:
+      facter: "3.13"
```
jbondpdx: This version gets Facter 3.13, but these docs are now integrated into Puppet docs in ezd. So I think maybe I can/should remove this line?

source/_config.yml
```diff
@@ -582,6 +590,19 @@ documents:
     version: "1"
     nav: /_includes/hiera1.html
 
+  /puppetserver/6.3:
+    doc: puppetserver
+    version: "6.3"
+    nav: /puppet/6.3/_puppet_toc.html
```
jbondpdx: This is Puppet Server 6.3, which is included in Puppet 6.4. If the version number on line 596 refers to a Puppet version, I'll change it to 6.4.

source/_config.yml
```diff
@@ -582,6 +590,19 @@ documents:
...
+    external_source:
+      repo: git://github.com/puppetlabs/puppetserver.git
+      commit: origin/master
+      subdirectory: documentation
+    my_versions:
+      puppet: "6.4"
+      facter: "3.13"
```
jbondpdx: Again, this version gets Facter 3.13, but those docs no longer come from GitHub or have their own URL. Should this line be removed?

source/_config.yml
```diff
@@ -582,6 +590,19 @@ documents:
...
+    hide: true
+
```
jbondpdx: Below, in the puppetserver/6.2 section, it lists puppet "6.2" under my_versions. This version of Puppet Server is relevant to two versions of Puppet: 6.2 and 6.3. Do I need to list puppetserver again, or add additional information here to reflect that this version is ALSO in Puppet 6.3?

jbondpdx:
Eh, I dug around some more and answered my own questions. Thank you though!

PR #923, opened 2019-04-15
(PDOC-272) Enable Requires features From Strings

kris-bosland:
With this change, Requires features data from strings parsing will show up in the types_strings output files.


jbondpdx:
Thank you, @kris-bosland !

PR #933, opened 2019-05-10
add note about backwards compatibility

nod0n:
https://tickets.puppetlabs.com/browse/DOCUMENT-982

### DIFF
--- a/source/puppet/6.4/configuration.md
+++ b/source/puppet/6.4/configuration.md
@@ -799,7 +799,7 @@ directories readable by Puppet Server when necessary.
 
 The hiera configuration file. Puppet only reads this file on startup, so you must restart the puppet master every time you edit it.
 
-- *Default*: $confdir/hiera.yaml. However, if a file exists at $codedir/hiera.yaml, Puppet uses that instead.
+- *Default*: $codedir/hiera.yaml. For backwards compatibility $confdir/hiera.yaml will be used if $codedir/hiera.yaml is missing.
 
 ### hostcert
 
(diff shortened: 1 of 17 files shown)


source/puppet/6.4/configuration.md
```diff
@@ -799,7 +799,7 @@ directories readable by Puppet Server when necessary.
 
 The hiera configuration file. Puppet only reads this file on startup, so you must restart the puppet master every time you edit it.
 
-- *Default*: $confdir/hiera.yaml. However, if a file exists at $codedir/hiera.yaml, Puppet uses that instead.
+- *Default*: $codedir/hiera.yaml. For backwards compatibility $confdir/hiera.yaml will be used if $codedir/hiera.yaml is missing.
```
jbondpdx: There's no change in 6.4 regarding codedir, so this should still be the same as the previous files:

*Default*: $confdir/hiera.yaml. However for backwards compatibility, if a file exists at $codedir/hiera.yaml, Puppet uses that instead.

nod0n: you are right. I don't know exactly how I messed that up. How ever, it's good you noticed it. :+1:

nod0n: how exactly do I accept your change?

nod0n: I'm not used to github and PR. Did I accept your change by clicking resolve conversation?

nod0n: now I think it's fixed. At least it looks like fixed for me in the "Files changed" tab.

PR #936, opened 2019-06-16
Fix incorrect method names in Facter::Util::IP

gene1wood:
It looks like there are no `get_ipaddress_value` and `get_netmask_value` methods in `Facter::Util::IP` and instead one should use the [`get_interface_value`](https://www.rubydoc.info/gems/facter/Facter/Util/IP#get_interface_value-class_method) method.

### DIFF
--- a/source/facter/2.4/fact_overview.markdown
+++ b/source/facter/2.4/fact_overview.markdown
@@ -158,8 +158,8 @@ Facter.add(:networking, :type => :aggregate) do
     interfaces = {}
     Facter::Util::IP.get_interfaces.each do |interface|
       interfaces[interface] = {
-        'ipaddress' => Facter::Util::IP.get_ipaddress_value(interface),
-        'netmask'   => Facter::Util::IP.get_netmask_value(interface),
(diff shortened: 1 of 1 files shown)


jbondpdx:
Hi @gene1wood ! First of all, I apologize that it's taken me so long to look at this PR!

I see that you made this to the Facter 2.4 docs, but I'm wondering if we need to make the same change in other versions. @branan , can you verify this fix and let me know if I should make the same change in current Facter versions?

gene1wood:
@jbondpdx No prob.

It looks like this example section exists in [3.0](https://github.com/puppetlabs/puppet-docs/blob/662c706ef18ea8850896639683bfceffb1b994c0/source/facter/3.0/fact_overview.md) but was removed in [3.1](https://github.com/puppetlabs/puppet-docs/blob/662c706ef18ea8850896639683bfceffb1b994c0/source/facter/3.1/fact_overview.md) and doesn't exist in 3.1 and newer versions. `3.0` is not present in the HEAD of master so I suspect the only current docs with this mistake are `2.4` which this PR corrects.

jbondpdx:
Thank you so much, @gene1wood !

PR #950, opened 2019-09-25
Remove versions

jbondpdx:
Heston, can you have a look at this PR? I'm removing a lot, so I just want a second set of eyes on it. 

The redirect commit for the Puppet Server and DB versions, if you want to refer to it, is at https://github.com/puppetlabs/picasso/pull/487

The Puppet redirects for <5.5 were merged a while back, but Server still depended on the old ToCs, so I'm just removing them now.

### DIFF
--- a/source/_config.yml
+++ b/source/_config.yml
@@ -625,28 +625,7 @@ documents:
     my_versions:
       puppet: "latest"
     hide: true
-  /puppetserver/6.5:
-    doc: puppetserver
-    version: "6.5"
-    nav: /puppet/6.7/_puppet_toc.html
-    external_source:
-      repo: git://github.com/puppetlabs/puppetserver.git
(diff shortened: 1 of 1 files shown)


hestonhoffman:
Okay I took a look at the redirects. The stuff you're removing looks good to me!  

A couple questions about stuff you _haven't_ removed. Forgive me if you've already told me this, I know we've had several conversations on this topic:

**Puppet 6.1, 6.2, 6.3** - This stuff is still in here, but it's not in the drop-down anymore on the site (Although you can still get to it using the URLs). Are you leaving it in cos there aren't redirects set up yet?

**Puppet 3.8** - Does that still need to be in there? Ancient long-lived version people are still using somewhere?

jbondpdx:
@hestonhoffman :
1. yes
2. sorry to say, yes

Thank you for looking at this!

PR #952, opened 2019-10-18
(maint) Add examples to function output

kris-bosland:
This updates the erb file that is used to create `references_output/puppet/function_strings_prefer_v4.md` so that examples are also output.


jbondpdx:
lgtm, thanks @kris-bosland !

PR #953, opened 2019-10-21
WIP, DO NOT MERGE

jbondpdx:
Changes to regsubst function for docs template work.

### DIFF
--- a/source/puppet/latest/function.md
+++ b/source/puppet/latest/function.md
@@ -3745,58 +3745,69 @@ $merged = $data.reduce( {} ) |$memo, $x| {
 
 ## `regsubst`
 
-* `regsubst(Variant[Array[String],String] $target, String $pattern, Variant[String,Hash[String,String]] $replacement, Optional[Optional[Pattern[/^[GEIM]*$/]]] $flags, Optional[Enum['N','E','S','U']] $encoding)`
-    * `target` --- The string or array of strings to operate on.  If an array, the replacement will be
+`regsubst(Variant[Array[String],String] $target, String $pattern, Variant[String,Hash[String,String]] $replacement, Optional[Optional[Pattern[/^[GEIM]*$/]]] $flags, Optional[Enum['N','E','S','U']] $encoding)`
+
+### Parameters
+
+* `target` --- The string or array of strings to operate on.  If an array, the replacement will be
 performed on each of the elements in the array, and the return value will be an array.
-    * `pattern` --- The regular expression matching the target string.  If you want it anchored at the start
+* `pattern` --- The regular expression matching the target string.  If you want it anchored at the start
 and or end of the string, you must do that with ^ and $ yourself.
-    * `replacement` --- Replacement string. Can contain backreferences to what was matched using \\0 (whole match),
+* `replacement` --- Replacement string. Can contain backreferences to what was matched using \\0 (whole match),
 \\1 (first set of parentheses), and so on.
 If the second argument is a Hash, and the matched text is one of its keys, the corresponding value is the replacement string.
-    * `flags` --- Optional. String of single letter flags for how the regexp is interpreted (E, I, and M cannot be used
+* `flags` --- Optional. String of single letter flags for how the regexp is interpreted (E, I, and M cannot be used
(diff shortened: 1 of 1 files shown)


jbondpdx:
@kris-bosland , here's a PR showing the template changes I'd like, on regsubst as the example. This is great progress, though!

source/puppet/latest/function.md
```diff
@@ -3745,58 +3745,69 @@ $merged = $data.reduce( {} ) |$memo, $x| {
...
+Get the third octet from the node's IP address:
+
+```puppet
 $i3 = regsubst($ipaddress,'^(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)$','\\3')
 ```
-    * Return type(s): `Array[String]`, `String`. The result of the substitution. Result type is the same as for the target parameter.
-* `regsubst(Variant[Array[String],String] $target, Variant[Regexp,Type[Regexp]] $pattern, Variant[String,Hash[String,String]] $replacement, Optional[Pattern[/^G?$/]] $flags)`
-    * `target` --- The string or array of strings to operate on.  If an array, the replacement will be
+
+Return type(s): `Array[String]`, `String`. The result of the substitution. Result type is the same as for the target parameter.
```
jbondpdx: The return type looks good, but should come before the example.

source/puppet/latest/function.md
```diff
@@ -3745,58 +3745,43 @@ $merged = $data.reduce( {} ) |$memo, $x| {
...
+* `encoding` --- Optional. How to handle multibyte characters when compiling the regexp (must not be used when pattern is a
 precompiled regexp). A single-character string with the following values:
   - *N*         None
   - *E*         EUC
   - *S*         SJIS
   - *U*         UTF-8
-    `Get the third octet from the node's IP address:`
-      ```puppet
+
+Return type(s): `Array[String]`, `String`. The result of the substitution. Result type is the same as for the target parameter.
```
jbondpdx: The return type should come before the examples.

source/puppet/latest/function.md
```diff
@@ -3745,58 +3745,43 @@ $merged = $data.reduce( {} ) |$memo, $x| {
...
+
+Get the third octet from the node's IP address:
+
+```puppet
 $i3 = regsubst($ipaddress,'^(\\d+)\\.(\\d+)\\.(\\d+)\\.(\\d+)$','\\3')
 ```
-    * Return type(s): `Array[String]`, `String`. The result of the substitution. Result type is the same as for the target parameter.
```
jbondpdx: All of this information is repeated before each dispatch call in regsubst.rb. So this is an error in the source, but if we delete the repeated information from the second dispatch call, does the second example still show up as it should?

source/puppet/latest/function.md
```diff
@@ -3745,58 +3745,69 @@ $merged = $data.reduce( {} ) |$memo, $x| {
...
 
-* `regsubst(Variant[Array[String],String] $target, String $pattern, Variant[String,Hash[String,String]] $replacement, Optional[Optional[Pattern[/^[GEIM]*$/]]] $flags, Optional[Enum['N','E','S','U']] $encoding)`
-    * `target` --- The string or array of strings to operate on.  If an array, the replacement will be
+`regsubst(Variant[Array[String],String] $target, String $pattern, Variant[String,Hash[String,String]] $replacement, Optional[Optional[Pattern[/^[GEIM]*$/]]] $flags, Optional[Enum['N','E','S','U']] $encoding)`
+
+### Parameters
+
+* `target` --- The string or array of strings to operate on.  If an array, the replacement will be
 performed on each of the elements in the array, and the return value will be an array.
-    * `pattern` --- The regular expression matching the target string.  If you want it anchored at the start
```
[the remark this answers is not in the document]
kris-bosland: *indented? :+1:

source/puppet/latest/function.md
kris-bosland: We should review what should be per-dispatch.

kris-bosland:
@jbondpdx , I will update my PR to fix these issues and then ask you for comment there.  Thanks!

PR #958, opened 2019-12-04
(DOC-3797) Remove `puppet cert` and "leftovers"

gguillotte:
The build scripts add "leftovers" -- `puppet` commands that aren't
explicitly listed as core, niche, etc. This was once a good idea;
now, thanks to `puppet cert`, it makes it impossible to complete
remove commands that have been disabled but still have scripts in
puppetlabs/puppet.

Remove the leftovers functionality and `puppet cert`. Any new
commands will need to be explicitly added to `core`, etc.
sections in `puppet-docs/lib/puppet_references/puppet/man.rb`.

### DIFF
--- a/lib/puppet_references/puppet/man.rb
+++ b/lib/puppet_references/puppet/man.rb
@@ -27,7 +27,6 @@ module PuppetReferences
             core: %w(
           agent
           apply
-          cert
           lookup
           master
           module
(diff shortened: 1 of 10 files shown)


gguillotte:
I think it's still valuable — people using 6 might still be looking here for `puppet cert`, so pointing them to the right place in the Server docs is still valuable.

jbondpdx:
Thank you so much @gguillotte !

PR #969, opened 2020-04-07
(DOCUMENT-1035): updated all document strings for reduce function

b4ldr:
A previous commit (20cd474) attempted to fix this issue however
only fixed the issue in the puppet 5.5 files.  This PR updates
the other files


jbondpdx:
Hi @b4ldr ! Thank you so much for the time you spent fixing this. Only 5.5 and the latest version of Puppet are currently maintained, which is why we didn't make the change throughout. None of the other versions are being published or updated.

This particular document is generated from strings in the codebase, so the latest version gets overwritten every release. Thanks to your PR, we discovered that the change to the codebase was never merged to master. I've manually made the change to the latest version of the docs, and we're PRing the fix to the codebase so that it doesn't pop back up later.

So I'm closing this PR, but we want to thank you again for your attention to this, which helped us fix the problem. The fix should be live after the next build of the docs this afternoon.

b4ldr:
@jbondpdx thanks for the update, it occurred to me after that the docs where probably generated.  

As a side comment and i appreciate this is definitely not the place to make this request (although i don't know where would be better).  It would be very useful if when closing tickets on tickets.puppetlabs.com a link to the appropriate pull requests is dropped in to the comments or somehow linked to the ticket, thanks

jbondpdx:
@b4ldr , the PRs are linked to the ticket, but I think this issue came up before, where those links are visible only to internal folks. I'm not sure why that should be the case on publicly visible tickets, so I'll look into the permissions on this. Thank you for raising it.

PR #985, opened 2020-07-10
Adding build for puppet 6.16 long-lived

hestonhoffman:
@jbondpdx I'm not 100% sure on this. I've never had to create a long-lived directory in puppet-docs for an LTS stream. What I did was copy the files from `latest` into a new directory - `6.16`, and added it to the build, so I have somewhere to point the build in kusama.


jbondpdx:
That's exactly right, @hestonhoffman . The one question I have is about whether we should add PuppetDB to the config file; if they are going to maintain a long-lived branch (which I would expect them to, on something going into PE LTS), then we should add it.

hestonhoffman:
Yeah, I'll check and add that. I need to double-check that server is going to cut that 6.12.x branch too

jbondpdx:
@clairecadman , if we maintain version-specific URLs for these PuppetDB and Puppet Server versions, I think you'll need to adjust any links within Puppet 6.16 to refer to the correct PuppetDB and Server versions instead of latest/

hestonhoffman:
**Update after talking to DB and Froyo**: DB and server won't have dedicated Y branches for this release for the time being. For now they're releasing Zs and using `master`. If they end up releasing something that's not intended to go into the LTS stream, they'll probably cut branches and we'll have to add them here. For now, we can just use `latest`.

PR #988, opened 2020-07-23
Version all gems, update rack

genebean:
Part of versioning all gems is the inclusion of the Gemfile.lock file and upgrading Dependabot.

This was tested by running the only task that is still in use (as I understand it): `body_and_nav_html_only`


genebean:
I am finished changing things now :)

PR #1047, opened 2021-02-10
(DOC-4857) Generate Facter CLI docs page

IrimieBogdan:
Added logic for generating Facter CLI documentation similar to the one for core facts. The same rake task that is used for core facts should be used, with the same arguments. The rake task will now generate `cli.md` in addition to `core_facts.md`.

If an older version of facter is used (older than 5.0.52) the CLI docs will not get generated and an error message will be displayed when running the rake task, but the core facts documentation will still be generated.

### DIFF
--- a/lib/puppet_references.rb
+++ b/lib/puppet_references.rb
@@ -22,6 +22,7 @@ module PuppetReferences
   require 'puppet_references/puppet/type_strings'
   require 'puppet_references/puppet/functions'
   require 'puppet_references/facter/core_facts'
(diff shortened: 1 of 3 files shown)


jbondpdx:
@hestonhoffman , this lgtm, please review and merge whenever you're ready.

jbondpdx (approved):
lgtm

hestonhoffman:
I couldn't get it to work off 5.0.52, but `main` worked fine. Looks good!

Thanks @IrimieBogdan!

PR #1064, opened 2021-04-28
Generate Puppet 7.6.1 and Facter 4.1.1 docs

clairecadman:
This commit generates the 7.6.1, without diffing the types.

### DIFF
--- a/source/puppet/latest/cli.md
+++ b/source/puppet/latest/cli.md
@@ -1,12 +1,12 @@
 ---
 layout: default
-built_from_commit: d3ee2f44b7cb0cadbb6fc9495f2eb093c4d6e148
+built_from_commit: b1ccc5d8d335a57108919a17a10d0ab80e91b2fc
 title: 'Facter: CLI'
 toc: columns
 canonical: "/puppet/latest/cli.html"
 ---
 
-# Facter CLI
+# TEST PREABLE
 
 ---
 SYNOPSIS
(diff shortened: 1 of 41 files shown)


source/puppet/latest/cli.md
```diff
@@ -1,12 +1,12 @@
...
 layout: default
-built_from_commit: d3ee2f44b7cb0cadbb6fc9495f2eb093c4d6e148
+built_from_commit: b1ccc5d8d335a57108919a17a10d0ab80e91b2fc
 title: 'Facter: CLI'
 toc: columns
 canonical: "/puppet/latest/cli.html"
 ---
 
-# Facter CLI
+# TEST PREABLE
```
clairecadman, line 9: Hi @IrimieBogdan! Could you please change the name of this in the source file? I can update it manually in the meantime.

clairecadman:
@hestonhoffman @jbondpdx This is my first time generating the docs without diffing the types. I think the type file differences look reliable, and it also pulled in Heston's fixes to the big tables. There are a couple of things that could be fixed in the source code for next time.

I'm not sure what is happening with `type_strings.json` though. Does that look right to you? Or should I not copy that one? Thanks!

source/puppet/latest/cli.md
clairecadman, line 9: Fixed.

hestonhoffman:
@clairecadman It looks like `type_strings.json` is pretty-printing in the new output. That should be fine. Mine looks the same.

hestonhoffman (approved):
LGTM. There are some changes in values from one version to the next (like changes in the listed providers), but I trust those are because of changes in the code.

